PR #684, opened 2013-10-11
Allow configuration location to be configurable

### LINKED ISSUE #610: v3: Possibility to move the config folder
Is it possible to add an option to specify the location of the config folder?

I see I can change the location of tasks folder if I edit the Capfile and replace this line:   

``` ruby
Dir.glob('lib/capistrano/tasks/*.cap').each { |r| import r }
```

In Capistrano 2 I had something like this in my Capfile:

``` ruby
set :stage_dir, 'deployment/deploy'
require "./deployment/railsless-deploy.rb"
load 'deployment/deploy' if respond_to?(:namespace)
```

seenmyfate:
This change allows both the `deploy_config_path` and `stage_config_path`
to be moved from the default locations of `config/deploy.rb` and
`config/deploy` respectively.  These values **must** be set in the
`Capfile` prior to `capistrano/setup` being called, for example:

```
set :deploy_config_path, 'app/config/deploy.rb'
set :stage_config_path, 'app/config/deploy'

# Load DSL and Setup Up Stages
require 'capistrano/setup'
```

Fixes #610

### DIFF
--- a/features/configuration.feature
+++ b/features/configuration.feature
@@ -1,0 +1,15 @@
+Feature: The path to the configuration can be changed, removing the need to
+  follow Ruby/Rails conventions
+
+  Background:
+    Given a test app with the default configuration
+    And servers with the roles app and web
+
+  Scenario: Deploying with configuration in default location
+    When I run "cap test"
+    Then the task is successful
+
+  Scenario: Deploying with configuration in a custom location
+    But the configuration is in a custom location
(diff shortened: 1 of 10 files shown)


webdevilopers:
If I'm getting this right the paths can only be changes after installing the Capfile?
Or is there an extra flag like `--config_path=...`?

PR #717, opened 2013-10-18
Introduce `no_release` server option

### LINKED ISSUE #686: no_release server option disappeared in v3
Capistrano v2 understood the `:no_release => true` flag to `server`, allowing tasks to execute on other servers without performing the normal release flow on them as well.

From my current reading of the v3 code, this is no longer the case. The use of `on roles :all` throughout the deploy flow blocks this behavior.

Is this option, or similar flexibility, going to come back? My environment is probably unique for Rails/Capistrano deploys - we deploy to a shared NFS filesystem, which means only 1 host needs to update the code or run migrations, but many hosts need to execute other actions within our flow.

I'm willing to create a pull request to make this happen, but I'd like a little guidance on the maintainer's preferred plan here. My current best thought is to change the

``` ruby
on roles :all do
  ...
end
```

blocks in the normal flow to something like:

``` ruby
on roles fetch(:deploy_role) do
  ...
end
```

along with adding `set :deploy_role, :all` to the default settings.

seenmyfate:
In order to provide a way for a server to perform tasks as part of a
deploy but without being involved in the standard deploy flow, all
included tasks will now prefer `release_roles` over `roles`.  For
example:

```
on release_roles :all do
  #
end
```

This behaviour is implemented using `exclude`, a new option when
selecting roles.  `release_roles` is equivalent to:

```
on roles :all, exclude: :no_release do
  #
end
```

Any server defined with `no_release: true` will be excluded from these
tasks:

```
server 'localhost', roles: %w{db}, no_release: true
```

`exclude` can also be used in user defined tasks against any attribute, for example:

```
server 'localhost', roles: %w{app web}, inactive: true

on roles :app, exclude: :inactive do
  #
end
```

This commit resolves #686


zacksiri:
I think we're still missing a few things like the clean up process when its trying to delete a release, on nodes that don't need release will be missing the releases/2023932923923 folder and throw an error

PR #756, opened 2013-11-01
Add hook for deploy failure

### LINKED ISSUE #708: Add Hook for Deploy Failure
When deployment fails, I need to call certain tasks (e.g. send a Hipchat notification). Previously this was accomplished via the on_rollback hook. This doesn't seem to exist in v3. Can it be added, or is there a task to which I can hook?

Edit: I don't need to rollback a specific task or even the entire deployment. I basically need to know when a deployment has failed, regardless of the specific task that failed.

seenmyfate:
If an error is raised during a deploy, the task `deploy:failed` will be
triggered.  Custom tasks can hook into this using `after`:

```
after 'deploy:failed', :send_for_help do
  #
end
```

I've also taken the opportunity to provide a marginally more useful
error message before triggering the task.

By default, this 'deploy:failed' will only be triggered when running
`cap <stage> deploy` - to trigger after individual tasks use `set
:deploying, true`

This closes #708 and replaces
https://github.com/capistrano/capistrano/pull/720

### DIFF
--- a/features/deploy_failure.feature
+++ b/features/deploy_failure.feature
@@ -1,0 +1,17 @@
+Feature: Deploy failure
+
+  Background:
+    Given a test app with the default configuration
+    And a custom task that will simulate a failure
+    And a custom task to run in the event of a failure
+    And servers with the roles app and web
+
+  Scenario: Triggering the custom task
+    When I run cap "deploy:starting"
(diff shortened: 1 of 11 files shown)


parkr:
Is this 3.x only? :frowning:

PR #822, opened 2013-12-06
Make Git submodules dance

coffeeaddict:
See #701 as to why.

(btw: I could be very blind, but there are no specs for `*.rake` files?)


DboyRedd:
coffeeaddict:git-submodules

PR #1041, opened 2014-05-05
Add Subversion authentication options

synewaves:
This builds off work started in PR 838 (#838) which added basic Subversion support. There were no options for authentication, which is used in most Subversion servers (usually when used over HTTP/S).

This adds two configuration variables:

:svn_username - Username
:svn_password - Password

When both are set, authentication is passed to the relevant Subversion commands.


leehambley:
This has been superseded by #1402.

PR #1046, opened 2014-05-07
Refactor enumeration of deployed releases

x3ro:
As mentioned in #994, our project structure requires folders other
than actual releases to exist in the releases_path. To be able to
customize the logic to enumerate releases (to exclude said non-
release folders), it has been refactored into a "configuration lambda"
`:releases`.

:warning: We haven't extensively tested this change yet. I hope that I didn't introduce any subtle errors, but you never know :smile: Hopefully some people can test this with their deployment to see if it breaks anything. The affected tasks should be `deploy:cleanup` (and thus implicitly `deploy`) and `deploy:rollback`.

I would write unit tests, but there doesn't seem to be a SSHKit backend suitable for that purpose at the moment. If there is and I just have missed it, please point me to it!

### DIFF
--- a/lib/capistrano/deploy.rb
+++ b/lib/capistrano/deploy.rb
@@ -1,3 +1,16 @@
 require 'capistrano/framework'
 
+# Fetches the currently deployed releases. Can be overwritten by setting :releases to a
+# different lambda.
+#
+# Example (inside a task)
+#
+#   releases = fetch(:releases).call(self)
+#   # => ["20140507122109", "20140507122237", "20140507122531"]
+#
+# Returns an array of folder names inside the releases_path, sorted oldest first.
+set :releases, lambda { |context|
+  context.capture(:ls, "-xt #{context.releases_path}").split.sort
+}
+
 load File.expand_path("../tasks/deploy.rake", __FILE__)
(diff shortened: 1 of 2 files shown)


Kriechi:
@leehambley is this already merged - or are you missing something?
I've rebased & updated my branch - should I open a second PR which closes this one?

x3ro:
@Kriechi Thanks for the rebasing! I've updated the branch to your state :)
@leehambley Can we :shipit: or should I close the PR?

leehambley:
Could ship it I think, I'll need to review when I'm at my PC.

x3ro:
:sparkles:

mattbrictson:
@x3ro Are you still around and up for continuing the discussion on this PR? I have some concerns that hopefully you can address and then we can merge this in.

1. `releases` is a very ambiguous term for something that has to provide very specific data. If for example a user overrides `releases` but does not sort correctly, they will be in for a nasty surprise the first time a cleanup or rollback occurs. I want to mitigate this scenario (which could involve data loss) if at all possible.
2. The default value for `releases` (or whatever it comes to be named) should go in `defaults.rb`, IMO.
3. Merge conflicts need to be resolved.

To guard against user error when defining `releases`, perhaps we can do something like:

* Change the name to `release_basenames_sorted` to make it clear we expect it to provide basenames (i.e. not complete paths) and sorted oldest to newest.
* Open a corresponding PR in the documentation to explain its usage.
* As a safety measure, make sure when cleaning up old releases, we check to make sure we aren't deleting the current release (which would happen if the sorting is incorrect).

mattbrictson:
Closing due to inactivity.

PR #1066, opened 2014-05-29
Apply HostFilter for fetch_primary

crhan:
When use capistrano with rails migration task, it will fetch the primary db server to run the migration. If I use HostFilter which does not include the primary db server. A "Directory Not Exists" error will occure.

So I think add the HostFilter to the `fetch_primary` method will help, and it does(for my situation).

## How to reproduce
```ruby
# deploy.rb
server "server1", roles: %w/app db web/
server "server2", roles: %w/app web/
```

### Deploy with hosts filter
```bash
cap production deploy --hosts=server2
```

### err logs
```
** Invoke deploy:migrate (first_time)
** Invoke deploy:set_rails_env
** Execute deploy:migrate
DEBUG [59ff49e4] Running /usr/bin/env if test ! -d /home/oob/apps/alioob/releases/20140528095735; then echo "Directory does not exist '/home/oob/apps/alioob/
releases/20140528095735'" 1>&2; false; fi on server1
DEBUG [59ff49e4] Command: if test ! -d /home/oob/apps/alioob/releases/20140528095735; then echo "Directory does not exist '/home/oob/apps/alioob/releases/201
40528095735'" 1>&2; false; fi
DEBUG [59ff49e4]        Directory does not exist '/home/oob/apps/alioob/releases/20140528095735'
```

The key point is db migration task on server1 is triggered, which is not as expected with the hostfilter.

### DIFF
--- a/lib/capistrano/configuration/servers.rb
+++ b/lib/capistrano/configuration/servers.rb
@@ -23,8 +23,9 @@ module Capistrano
       end
 
       def fetch_primary(role)
-        hosts = roles_for([role])
-        hosts.find(&:primary) || hosts.first
+        hosts = fetch([role])
+        primary_host = hosts.find(&:primary) || hosts.first
+        HostFilter.for([primary_host]).first
       end
 
       def each
--- a/lib/capistrano/configuration/servers.rb
+++ b/lib/capistrano/configuration/servers.rb
@@ -23,9 +23,8 @@ module Capistrano
       end
 
       def fetch_primary(role)
-        hosts = fetch([role])
-        primary_host = hosts.find(&:primary) || hosts.first
-        HostFilter.for([primary_host]).first
+        hosts = roles_for([role])
+        hosts.find(&:primary) || hosts.first
       end
 
       def each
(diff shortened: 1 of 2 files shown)


crhan:
@leehambley rebased

crhan:
but wait....the HostFilter code has changed a lot.... I'm fixing it

crhan:
It's odd, the thing I want to prevent is the thing that @townsen think is correct. (Referencing the commit https://github.com/capistrano/capistrano/commit/ede09c93e80155716ebf2ef0f2e7084a1521b55e)

What do you think for this problem? @townsen 

What I'm going to do is to filter out the server that _dose not_ have the primary flag. Aiming to prevent database migration in rails deployment. (In the case that deploy a new server, so I the primary db server should be filtered out)

crhan:
hummmm, It seems like problem is fixed by @townsen at https://github.com/capistrano/capistrano/commit/ede09c93e80155716ebf2ef0f2e7084a1521b55e

Is this problem really exist? @joshmyers please check the spec I add in this PR

PR #1076, opened 2014-06-08
Use correct capistrano version in deploy.rb.erb

ojab:
Use actual capistrano version instead of hardcoded '3.1'.

### DIFF
--- a/lib/capistrano/templates/deploy.rb.erb
+++ b/lib/capistrano/templates/deploy.rb.erb
@@ -1,4 +1,4 @@
-# config valid only for Capistrano 3.1
+# config valid only for Capistrano <%= Capistrano::VERSION %>
 lock '<%= Capistrano::VERSION %>'
 
 set :application, 'my_app_name'


Kriechi:
Can we detect if the generator is called from a Bundler-environment?
If yes I would suggest only adding the version-lock in generator calls from non-bundler environments.

jgrevich:
@Kriechi perhaps just checking for the prescence of a `Gemfile` or `Gemfile.lock

Kriechi:
@jgrevich just looking for a `Gemfile` is probably a bad idea.
But to keep it simple I would be happy with just changing the comment to something like:


```ruby
# If you do not use Bundler to manage your Gem versions, 
# you can lock this Capistrano config to a specific version.
# lock '<%= Capistrano::VERSION %>'
```

Please note that the `lock` command is commented out by default, I think most of the people use Capistrano in a Bundler-env - so I would not want to bother these guys with version locking.

jgrevich:
> Please note that the lock command is commented out by default, I think most of the people use Capistrano in a Bundler-env - so I would not want to bother these guys with version locking.

:+1:

kirs:
Thanks for your PR.
This is already fixed in https://github.com/capistrano/capistrano/blob/master/lib/capistrano/templates/deploy.rb.erb#L1

PR #1109, opened 2014-08-01
Replaced Hash#has_key? is deprecated method with Hash#key?

justanshulsharma:
http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/43765

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,7 @@ require "bundler/gem_tasks"
 require "cucumber/rake/task"
 require "rspec/core/rake_task"
 
-task :default => :spec
+task default: :spec
 RSpec::Core::RakeTask.new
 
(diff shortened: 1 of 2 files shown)


Kriechi:
@justanshulsharma seems you missed one (or it was added recently):
https://github.com/capistrano/capistrano/blob/239b03a7013677fcbaac3ea8ba0f95cdd8ab4de6/lib/capistrano/configuration.rb#L31

(I grepped the latest master as of commit 239b03a70)

leehambley:
Thanks for the commit, unfortunately the patch is totally out of date (my fault) I have performed the same change by hand, and everything seems fine. Thanks again (I credited you in the README, unfortunately not in the git commit itself.)

PR #1187, opened 2014-11-16
Make relative instead of absolute links

hoesler:
Links within the deploy_to folder should be relative instead of absolute, which makes it possible to move the whole deploy_to folder.

### DIFF
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -83,7 +83,7 @@ namespace :deploy do
     task :release do
       on release_roles :all do
         execute :rm, '-rf', current_path
-        execute :ln, '-s', release_path, current_path
+        execute :ln, '-s', release_path.relative_path_from(current_path.dirname), current_path
       end
     end
 
@@ -106,7 +106,7 @@ namespace :deploy do
             if test "[ -d #{target} ]"
               execute :rm, '-rf', target
             end
-            execute :ln, '-s', source, target
+            execute :ln, '-s', source.relative_path_from(target.dirname), target
           end
         end
       end
@@ -125,7 +125,7 @@ namespace :deploy do
             if test "[ -f #{target} ]"
               execute :rm, target
             end
-            execute :ln, '-s', source, target
+            execute :ln, '-s', source.relative_path_from(target.dirname), target
           end
         end
       end


leehambley:
Closing as you were able to work around this, and I spoke with a colleague who's a bit more of a UNIX beard than I am who agreed that the current behaviour is probably correct, even if sometimes it can be inconvenient (but at least it's easy to patch)

gagarine:
Would be nice to have a ready example. I asked the question on http://stackoverflow.com/questions/31590392/capistrano-3-relative-symlink-instead-of-absolute-for-current-linked-dirs-and

gagarine:
Oh sorry I skipped the patch... I will try to use that.

x-yuri:
> who agreed that the current behaviour is probably correct

What do you mean by "correct"? I have a server with LXC containers. I want to obtain Let's Encrypt's SSL certificate for one of them. But absolute symlinks in a container doesn't work on a physical machine. Relative symlinks would solve the issue.

leehambley:
>What do you mean by "correct"? 

Our definition of correct is that it works for the typical Capistrano use-case, and that it's in line with what most of our users would expect.

> I have a server with LXC containers. I want to obtain Let's Encrypt's SSL certificate for one of them. But absolute symlinks in a container doesn't work on a physical machine.

Sounds like a problem with the way you're using LXC. I don't see how absolute path symlinks would help.

If it's so much of a problem for your niche use-case, then please read the docs http://capistranorb.com/documentation/advanced-features/overriding-capistrano-tasks/

x-yuri:
Don't take me wrong. I'm not trying to criticize you. I'm just trying to make `capistrano` a better "place."

The thing is, absolute symlinks doesn't work outside of any chroot. As for relative symlinks, nothing comes to mind. I've inspected the history of the repo, and seen no traces of symlinks ever being relative. I read about `mod_realdoc` module. And see no relevance to the issue. To sum it up, we know what issues can cause absolute symlinks, but no issues known for relative ones.

But sure, it's up to you to decide.

leehambley:
> But sure, it's up to you to decide.

The decision was made by the whole community two years ago!

I've linked you our official docs on how you can make a ~4 line change to suit your use-case.

x-yuri:
I'm well aware how to override a task. Like I said, my point is to make `capistrano` better. Things like this end up cluttering files, making it hard to see what's important. To me, it's supposed to work out of the box. But if community thinks different [for no apparent reason][1]... there's nothing I can do, is there?

[1]: https://stackoverflow.com/questions/44541937/are-there-any-advantages-to-using-absolute-symlinks-over-relative-ones

ain:
My 2 cents on the topic: a decision that is 2 years old, at its current pace, is probably already obsolete. On that note Rails 5.1 just dropped jQuery.

I guess I'm following this topic over `certbot` topic that I tackled with on HAProxy side as well, but I  solved it with a separate backend definition rather than on Cap side. The important matter is always to solve the problem in its scope and I didn't see it in Cap scope back in the days. It's a `cerbot` and server routing problem, not a Cap problem.

leehambley:
Case closed. Thanks.

PR #1301, opened 2015-01-09
Disable default extended server syntax in template

ain:
- Comment out extended syntax
- Add attention notice as part of recommendation

### DIFF
--- a/lib/capistrano/templates/stage.rb.erb
+++ b/lib/capistrano/templates/stage.rb.erb
@@ -3,6 +3,7 @@
 # Supports bulk-adding hosts to roles, the primary server in each group
 # is considered to be the first unless any hosts have the primary
 # property set.  Don't declare `role :all`, it's a meta role.
+# ATTENTION! It is recommended to use single syntax, either simple or extended.
 
 role :app, %w{deploy@example.com}
 role :web, %w{deploy@example.com}
@@ -14,8 +15,9 @@ role :db,  %w{deploy@example.com}
 # This can be used to drop a more detailed server definition into the
 # server list. The second argument is a, or duck-types, Hash and is
 # used to set extended properties on the server.
+# ATTENTION! It is recommended to use single syntax, either simple or extended.
 
-server 'example.com', user: 'deploy', roles: %w{web app}, my_property: :my_value
+# server 'example.com', user: 'deploy', roles: %w{web app}, my_property: :my_value
 
 
 # Custom SSH Options


Kriechi:
I am in favour of changing *something*.
But in my opinion the **extended** syntax is the simpler one and should be kept enabled.
The **simpler** syntax should be commented out IMHO because it takes at least 3 three lines of config, where as with the **extended** syntax everything can be done in just one (still easy to read) config line.

Without any hard evidence, but from my personal experience most deployments out there are doing web+app+db on a single host.

Not sure if it is worth the effort of relabelling the two syntaxes or am I the only one feeling it should be swapped?

@kirs @leehambley any suggestions on that?

ain:
Thomas has a point of course, but we should consider a best practice or a
recommendation by the Capistrano community. I'd also leave it with @kirs
and @leehambley to decide :)

kirs:
I see no point in providing 2 ways and recommending only one.

ain:
The problem is, if both are in by default, you could possibly have
conflicts when leaving both in…

I had both on on one of the projects until @leehambley pointed it out. One
of the recent releases must have changed the behaviour, because I've kept
both on on projects that have been running for severals years.


-- 
Sent from Gmail Mobile

kirs:
What about mentioning that using both styles can lead to conflict?

ain:
I appended a short line on that to comments above on both of them :)

townsen:
I think it is important to have both the role-based (_simple_) and server-based (_extended_) definitions. On many of my projects we wish to give all the servers in a role the *same* properties, except perhaps for one specific server that has something different or extra. To keep things as DRY as possible you begin with a role declaration that sets the parameters for all the servers, and then follow it with a specific server declaration that updates only the changed one. Here's a simple contrived example:
```ruby
role :redis, %w{server1 server2 server3 server4}, redis: { port: 6379, slave: true}
server 'server1', redis: { slave: false }
```
This works in later versions of Cap, but didn't historically. I submitted a PR some considerable time ago that _merged_ properties when they are defined more than once. Prior to that the second declaration would obliterate the first. I did this as it fits with what you might expect to happen. So when you say that using both styles can lead to conflict - is this what you mean? If so then I'd suggest we just describe this merge process in the template and recommend that people use whatever meets their needs. I don't see this as a 'conflict' - more like a 'feature'!

The merge process is essentially:
* Scalar properties are over-ridden, duplicate names = last one wins
* Array properties are appended to
* Hashes are merged just at the first level, duplicate keys = last one wins

Kriechi:
Sounds like a valid use case at first glance.

But I would argue it to be "advanced" - and therefore I would not include it into the template.
Maybe a well written documentation page would be better suited for this merging and overwriting of settings stuff...

leehambley:
We're keeping both syntaxes, it's a bug we have at the moment where properties (and, server comparison) behaviour has changed, and that we incorrectly merge them, people obviously need the combined server syntaxes, and there's no reason not to just fix it when we have the time/capacity!

Kriechi:
@leehambley do you want to keep both syntaxes in the template?

I would opt for only showing the "simple" syntax. And maybe a two-line comment mentioning a docs page with the advanced syntax.

leehambley:
I think that the "advanced" syntax is common enoughly used that it's not "advanced and simple" rather "a and b"

Kriechi:
either way - it's confusing the have both active in the template...
Which leads us back to the original topic of this PR.

I do use only the *advanced* syntax, because all I need is a single config line.
As mentioned above **simple**, **advanced** or **extended** are misleading names.

How about renaming them to something like
* Role-based syntax
* Server-based syntax

and not enabling anything in the template, but put everything as comment?

ain:
Yup, Thomas got me right in his last comment. That was the idea behind the
ticket.

I'd say better comments are necessary and renaming syntaxes is also good,
but I'm mixed about commenting both out. We do actually have lines in from
scratch elsewhere. It'd be a somewhat inconsistent to have both syntaxes
off by default here?

leehambley:
Alright, so we agree: 

* Role-based syntax
* Server-based syntax

annnddd, what do we do about it? Does anyone of you know already if master (we have some open PRs too from @Kriechi and others) about fixing host properties, and filtering, and we have recent patches from @FooBarWidget fixing server/host comparison… I'd ideally like to get it fixed, and I'll kick myself if there's a PR I missed that fixes things.

leehambley:
@ain @Kriechi I don't have time before Thurs 05.02.15 to look at cap in any useful way, if you could do anything to cleanup the naming of this stuff I would be very grateful.

townsen:
I've been away on vacation most of January, but now have time to look at some of these PR's and problems/issues with host and role properties. Can you point out the things you think need attention? (I'll go look over the PR's and issues but it would help if you could point out the most important ones). I think better doc will also help people understand what's going on - so I'll document the property merging (which I don't think is easily available - or is it?)

leehambley:
@townsen property merging only as part of a pending PR afaik.

I think first I need to understand what I expect the behaviour to be! (additive, always additive, I guess… but then the question becomes… added to *what*, purely hostname, or do we take the username into account?)

PR #1321, opened 2015-01-12
Do not suggest `bin/` as a shared directory by default

kirs:
Two reasons:
1) `bin/` is often checked out into repo
2) https://github.com/capistrano/bundler/issues/45#issuecomment-69349237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Reverse Chronological Order:
 ## master
 * Fixed updating roles when custom user or port is specified. (@ayastreb)
 
+* `bin/` is not suggested to be in `linked_dirs` anymore (@kirs)
+  * bin/ is often checked out into repo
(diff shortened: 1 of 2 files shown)


Kriechi:
:+1:  for removing `bin/` 

I would suggest adding `config/secrets.yml` to `linked_files` - which is commented out, but at least add it to the comment.

Reason: `config/database.yml` is already in there, and *feels* Rails-related. So it would make sense to add `config/secrets.yml` as well to minimise confusion for Rails users. I'm not talking about setting it as default, just add it to the commented line in the template https://github.com/capistrano/capistrano/pull/1321/files#diff-348edd3e734eaeed0fe97af037f3c647R26.

kirs:
@Kriechi great idea. Could you open a PR, please?

Kriechi:
@kirs done, see #1337.

PR #1337, opened 2015-01-18
add secrets.yml to linked_files in deploy.rb

Kriechi:
Currently linked_files already contains config/database.yml, which 
could be considered a Rails-like configuration default.

Adding config/secrets.yml streamlines the recommended configuration for
Rails 4.1+ applications.

see #1321 which does a similar thing to `linked_dirs` for the same reasons.


kirs:
Thanks!

sponomarev:
maybe capistrano-rails is better place to store so specific configuration?

Kriechi:
It's just a default in a non-active config line.
capistrano-rails can't change the default template generator...

PR #1338, opened 2015-01-18
cleanup readme

### LINKED ISSUE #1336: Move large parts of README to capistranorb.com
Currently most parts of documentation are kinda duplicated or (only in one place) in the README.md and the capistranorb.com website. Ideally there should only be one single documentation, in one place, updated every time a change happens or features get added. Currently one has to check both places and keep them in sync.

I propose:
moving large chunks out of the `README` and 
into the website repo at https://github.com/capistrano/capistrano.github.io. 

The `README` **should contain**: 
* Requirements (and VCS requirements)
* Installation
* Usage
* Testing
* SSHKit
* License

The following sections **currently present** in the `README` **should be moved** to the website repo:
* Tasks
* Before / After
* Getting user input
* Password authentication
* Local tasks
* Console
* PTYs
* Configuration
* Host and role filtering
* Metrics

I'm willing to create two PRs, one for cleanup here, one for adding missing stuff to the website repo.
/cc @leehambley @seenmyfate

Kriechi:
closes #1336 

Please see #1336 for reasons and explanation.
This PR is accompanied by a second PR for the documentation repository, which more or less adds the parts, deleted in this PR from the README, to the web documentation.
https://github.com/capistrano/capistrano.github.io/pull/110

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,26 @@
 # Capistrano [![Build Status](https://travis-ci.org/capistrano/capistrano.svg?branch=master)](https://travis-ci.org/capistrano/capistrano) [![Code Climate](http://img.shields.io/codeclimate/github/capistrano/capistrano.svg)](https://codeclimate.com/github/capistrano/capistrano) <a href="http://codersclan.net/?repo_id=325&source=small"><img src="http://img.shields.io/badge/get-support-blue.svg"></a>
 
-## Requirements
+## Documentation
 
-* Ruby >= 1.9.3 (JRuby and C-Ruby/YARV are supported)
+Check out the [online documentation](http://capistranorb.com) of Capistrano 3 hosted via this [repository](https://github.com/capistrano/capistrano.github.io).
 
 ## Support
 
 Need help with getting Capistrano up and running? Got a code problem you want to get solved quickly?
 
-Get <a href="http://codersclan.net/?repo_id=325&source=link">Capistrano support on CodersClan.</a>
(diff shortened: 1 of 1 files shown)


kirs:
Actually, we're going to have a problem: when changing public APIs we won't be able to update the docs in the single commit

leehambley:
@kirs I considered that too, actually having a README was one of the primary reasons for keeping docs in line with the code, so you could check both out at a specific tag. If our inter-release APIs start to get too confusing, we'll have to look at versioning the capistranorb.com site, the same way that Rails API browsers (and friends) are split by release version.

I hope we never come to that :)… and besides, we never break public APIs, right :-D

Kriechi:
Or we could merge the two repos into a single one (e.g. merging the docs into capistrano).
But I don't know how the GitHub Pages stuff works - redirects, branches, etc...

leehambley:
> But I don't know how the GitHub Pages stuff works - redirects, branches, etc...

Typically I tend to start an [orphan branch](http://git-scm.com/docs/git-checkout/1.7.3.1) onto which the `gh-pages` stuff goes. The histories are then distinct and unique. However it's not super discoverable, that said, we don't get a lot of doc contributions from outside the de-facto core team anyway, so perhaps we already have a discoverability problem/

Kriechi:
well, using a `gh-pages` branch still does not solve the "single commit API change" problem.
Ideally the docs would be in capistrano/master. And some outside entity pushes the stuff to `capistranorb.com`.

leehambley:
Perhaps we can try something with a CI service to copy everything from `master` in a `gh-pages` folder to the root of a `gh-pages` orphan branch, and push the result.

Kriechi:
Jep - thought about that as well. ZNC are doing that to generate their API docs. Everytime travis runs on master is compiles the docs und pushes them to the docs repository: https://github.com/znc/docs

leehambley:
Interesting :-\ … maybe it's worth doing that.

PR #1339, opened 2015-01-20
Fixed fetch revision for annotated git tags

igorsokolov:
Hello!

Our deployment process is using git tags to keep all apps in sync. There is an issue:

When you create git tag with annotation by doing : 
```
git tag -a <tag-name> -m "Message"
```
it creates new object with message. So 
```
git rev-parse <tag-name> 
```
returns the commit SHA of that object, while 
```
git rev-list --max-count=1 <tag-name>
```
returns the tagged commit SH A.

--abbrev-commit is for short commit SHA

So this change helps to keep track on correct revision when :branch is set to annotated tag.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 Reverse Chronological Order:
 
 ## master
+* Fixed fetch revision for annotated git tags. (@igorsokolov)
 * Fixed updating roles when custom user or port is specified. (@ayastreb)
 
 * `bin/` is not suggested to be in `linked_dirs` anymore (@kirs)
--- a/lib/capistrano/git.rb
+++ b/lib/capistrano/git.rb
@@ -40,7 +40,7 @@ class Capistrano::Git < Capistrano::SCM
     end
 
     def fetch_revision
-      context.capture(:git, "rev-parse --short #{fetch(:branch)}")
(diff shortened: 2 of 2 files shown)


leehambley:
Seems to work for me on normal tags and branches too, thanks for the contribution, I wonder if this will fix some of the reported issues around the revision list being incorrectly saved (under poorly defined bug report conditions!)

igorsokolov:
Thank you ! 

Any plans on releasing new version anytime soon ?

leehambley:
Releasing a new version has huge implications for deep testing, so it's rare that I can make the time for it. Especially tricky at the moment as my family and I are moving homes at the end of January, I hope it won't take that long, but I need to set aside half a day to get the next release out. I'd suggest just using `master` from your `Gemfile` in the meantime.

igorsokolov:
Great! Just wanted some hints on this.
Thanks !

igorsokolov:
Hi ! 
We are still pointing to `master` branch while waiting for a new version released. And now there is an issue: 

https://github.com/capistrano/capistrano/blob/master/lib/capistrano/deploy.rb#L2

requires deleted dependency. 

Thanks!

leehambley:
Fixed, you could also have added `capistrano-stats` to your own gemfile.

Issue #1341, opened 2015-01-20
Roles aren't matching explicitly.


sammarx:
I have 2 roles that have similar names.
config/deploy/test.rb
```
role :test, "test-server-1"
role :asdftest2, "test2-server-2"
puts roles(:test)
```
Test task, just to execute our stage
```
task :test do
end
```
Run cap, should only output "test-server-1"
```
$cap test test
test-server-1
test2-server-2
```

It looks like Cap is matching any part of the of the string, instead of the string as a whole.

Kriechi:
#1342 got merged and fixes this issue.
@leehambley please close this.

PR #1342, opened 2015-01-20
Fixed regex matching for roles to be explicit

sammarx:
Addresses https://github.com/capistrano/capistrano/issues/1341


leehambley:
I already merged, since you didn't adhere to the CONTRIBUTING guidelines, I'l assume you didn't want to be credited in the Changelog. Thanks for your patch.

sammarx:
Works for me. Thanks for the quick turnaround!

PR #1350, opened 2015-01-27
Mark version 3.3.3's tracking of anonymous statistics as a breaking change

Narnach:
After upgrading from 3.2.x to 3.3.x and scanning the changelog for breaking changes, this change broke automated deployment via our continuous deployment. I had to run a manual deploy to create the settings file, after which our CI server could deploy again.

I'm baffled that this was not marked as a breaking change, but I hope to rectify it after the fact with this commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,7 +58,6 @@ https://github.com/capistrano/capistrano/compare/v3.2.1...v3.3.3
     and --trace options supply it as before.
 
 * Disable loading stages configs on `cap -T` (@sponomarev)
-* Track (anonymous) statistics, see https://github.com/capistrano/stats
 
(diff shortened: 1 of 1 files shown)


leehambley:
Thanks for the PR, but the changelog doesn't change retrospectively.

The statistics Gem goes away soon, and we tweeted and posted to the mailing list about it! 

I'm glad to see someone reads the CHANGELOG, though.

Unfortunately tracking statistics, and badly behaved CIs (faking a TTY, and failing to respond to input, etc) has caused too many problems for people, so we're removing the requirement imminently.

leehambley:
Actually, why not… if it saves one person three minutes, then it's your time well invested. Thanks.

Narnach:
Thank you!

PR #1351, opened 2015-01-27
Release roles bug

townsen:
The `release_roles()` function did not handle additional selection options correctly - changing `dsl_spec.rb:18` highlighted the problem with the existing code. I reworked the Property filtering code to make it cleaner and more capable (added shortcut for property equality tests).

Improved documentation and added new tests


townsen:
Will resubmit this after sync'ing up with latest master....

leehambley:
Thanks Nick

PR #1352, opened 2015-01-27
Release roles bug - improve property filtering

townsen:
Property filtering did not honour multiple filter options. Highlighted by changing `dsl_spec.rb:18`. This PR fixes that, cleans up the property filter code and adds a new capability (testing server property equality). I'll PR some documentation to capistrano.github.io to describe this.


leehambley:
Thanks Nick

PR #1354, opened 2015-01-28
Correcting typo in docs

kmwhite:
Correcting a typo of 'come' -> 'some' when describing usage of Capistrano from a non-tty source.


leehambley:
Thanks :) recent addition to the docs there from a PR from a non native speaker, nice catch and thanks again!

PR #1357, opened 2015-02-03
Set repo path variable - update Vagrantfile and spec tests

townsen:
This PR adds one small piece of functionality and tidies up some tests:
* Allow specification of `:repo_path` to set where the repo is cloned
* Move all the Paths tests out of dsl_spec into paths_spec
* Updates the Vagrantfile so the features run with Vagrant 1.7+ (this may require devs to upgrade from older versions)


leehambley:
Thanks @townsen

PR #1368, opened 2015-02-14
Consolidate servers on hostname

townsen:
This PR contains the changes we discussed about merging servers with the same hostname
(but potentially differing users and ports).

I also enhanced the `on()` method to use a copy of the server object.

See documentation in the PR to capistrano.github.io for more details

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ Reverse Chronological Order:
   * release_roles did not honour additional property filtering (@townsen)
   * Refactored and simplified property filtering code (@townsen)
 
+* Breaking Changes
+  * Hosts with the same name are now consolidated into one irrespective of the
+    user and port. This allows multiple declarations of a server to be made safely.
+    The last declared properties will win. See capistrnorb.com Properties documentation
+    for details.
+  * Inside the on() block the host variable is now a copy of the host, so changes can be
+    made within the block (such as dynamically overriding the user) that will not persist.
+    This is very convenient for switching the SSH user temporarily to 'root' for example.
+
 * Minor changes
   * Add role_properties() method (see capistrano.github.io PR for doc) (@townsen)
   * Add equality syntax ( eg. port: 1234) for property filtering (@townsen)
--- a/features/sshconnect.feature
+++ b/features/sshconnect.feature
@@ -1,0 +1,17 @@
+Feature: SSH Connection
+
+  Background:
+    Given a test app with the default configuration
+    And servers with the roles app and web
+    And a task which executes as root
+
+  Scenario: Switching from default user to root and back again
+    When I run cap "git:check"
+    Then the task is successful
+    And references in the remote repo are listed
+    When I run cap "am_i_root"
+    Then the task is successful
+    And contains "root" in the output
+    Then I run cap "git:check"
+    Then the task is successful
+    And references in the remote repo are listed
--- a/features/step_definitions/setup.rb
+++ b/features/step_definitions/setup.rb
@@ -27,6 +27,10 @@ Given(/^a custom task to generate a file$/) do
   TestApp.copy_task_to_test_app('spec/support/tasks/database.rake')
 end
 
+Given(/^a task which executes as root$/) do
+  TestApp.copy_task_to_test_app('spec/support/tasks/root.rake')
+end
+
 Given(/config stage file has line "(.*?)"/) do |line|
   TestApp.append_to_deploy_file(line)
 end
--- a/lib/capistrano/configuration/server.rb
+++ b/lib/capistrano/configuration/server.rb
@@ -62,7 +62,7 @@ module Capistrano
       end
 
       def matches?(other)
-        user == other.user && hostname == other.hostname && port == other.port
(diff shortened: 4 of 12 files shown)


Kriechi:
@townsen why did you introduce the possibility to do
```ruby
host.user = 'root'
```

Isn't that already covered with the SSHKit `as 'root'` block?
https://github.com/capistrano/sshkit/blob/master/EXAMPLES.md#run-a-command-as-a-different-user
Or am I missing a point here?

townsen:
The short answer is to provide finer-grained access control within tasks.

We have deployment scenarios where we want the user SSH'ing in as the deployer _not_ to generally have root access, but we have tasks that must run as root. The `as()` block uses `sudo` to switch to root and if you give this power to the deploy user then *all* users (with a deploy user key) will have root access.

By setting the SSH connection to use 'root', access is determined by the `authorized_keys` file for root. This is separate to that for the deploy user (though people may have keys in both!). Everyone can _try_ and run every task, but only those with the correct keys in place will actually succeed.

This is more flexible than the sudo method. I'd also thought of re-implementing this as an `as()` style block construct but that would confuse people I think, so felt it was best kept as an SSH property.

Thoughts?

leehambley:
It'll need clear documentation, I've never had a problem, between setuid scripts, proper groups, limiting what users can do using the SSH config, and per-command sudo access .(alright, all together might be hell, but still the point stands).

Of course there's also AppArmor and SELinux, these are neat tools to stop users wandering outside their "I need sudo to install gems and reboot" remit.

leehambley:
I've always tried to walk the line between documenting, or not documenting what, actually has very little to do with Capistrano. I tipped the balance in favour of "document more of how the underlying system" works, intentionally with Cap3. (Cases in point, the gateway config: use `.ssh/config`, and the PTY and "why does something work in SSH but not in cap" docs I added)

This is also in that vein, in my opinion.

Kriechi:
@townsen ok - I think I got it now. Thanks. Valid use case I guess.

**IMHO**
I don't like the idea of spreading authentication and authorisation over two different mechanisms (`sudo`-permissions and actual SSH user). 
Additionally I think allowing general root login via SSH is a bad idea. Special use cases might justify it though...

As always in *nix there are many different paths to the same goal. Some are shorter, some provide a nice view, and others should only be used with closed eyes while praying.

townsen:
I agree with keeping the documentation (and implementation) simple so it's best to leave things as they are (no new `as()` functionality). As you point out the merits and methods of SSH vs sudo are well covered elsewhere. This just lets people choose either method easily.

Note that although I used 'root' in the example, it's useful for non-root users too. We use a _warden_ user to handle secrets of various sorts, with different tasks and permissions to the deploy user on certain machines.

bibendi:
:hankey: You broke us all deploys!

leehambley:
> :hankey: You broke us all deploys!

Of course you tested Capistrano in your staging environment before deploying to production. At least it wasn't serious!

maxim:
Essentially what this does is render roles at best redundant, and at worst obsolete. I have deployer and runner users. Runner can do things like read logs and launch remote console, while deployer can only deploy files. So now I can't declare the same host twice, once in logging + code execution role with runner user, another with deployer user. I would have to list all the servers I need to get logs from in .ssh/config, which means I have to list all the servers multiple times in different places. Is there a workaround for this?

maxim:
Hm I think I finally figured out how it's intended to be used. The new way is to set default user in your stage, and any alternative user should be set as needed in individual `on` blocks. I think this can work given we also have ruby.

townsen:
@maxim,
Yes - you've got it. You can now just set the user in the `on()` block like so:
```ruby
      on docker_host do |host|
        host.user = 'root'
        host.ssh_options = { proxy: nil }
        execute :chef_solo, '-o', "recipe[c-pod::docker_host]"
      end
```
I don't consider this a workaround - I consider it a *feature*, and it also keeps the configuration very DRY.

Regards
Nick

leehambley:
I don't consider it a *workaround* either, but maybe we should document it somewhere Nick as I wouldn't have thought of this, I always go for the "give it a name that represents it's role, and define the properties in the local ssh config". I think the mindset depends a lot if you think of servers in terms of their roles, or their hostnames, maybe it's related to how warped our minds are from CM tools :)

betesh:
>    on docker_host do |host|
>        host.user = 'root'
>        host.ssh_options = { proxy: nil }

Workaround or now, it's only usable in user-defined tasks.  Many tasks are defined in gems.

In my case, I have some cron jobs, some of which run as the deployment user, and some of which need to run as a different user.  Using the whenever gem, this used to work:

    set :whenever_roles, -> { [:my_host, :my_host_as_other_user] }

And then, in config/schedule:

    every :day, roles: [:my_host] do
      command :script1
    end

    every :day, roles: [:my_host_as_other_user] do
      command :script2
    end

In summary, this change forces a complete re-write of [whenever's capistrano tasks](https://github.com/javan/whenever/blob/v0.9.4/lib/whenever/capistrano/v3/tasks/whenever.rake)

PR #1369, opened 2015-02-14
Improve Cucumber tests

townsen:
Although the cucumber test feature was passing - it wasn't ever going to fail!! This is what happens when you write the test afterwards. My bad.

Having realized this I fixed the feature and tidied up the regexes too.


leehambley:
Cheers @townsen I didn't notice it either!

PR #1386, opened 2015-03-03
remove post-install message

Kriechi:
This PR removes the post_install_message from `gemspec`.

v3.0 and v3.1 are both over a year old (Oct-2013 and Jan-2014).
IMHO the message is not needed anymore.


leehambley:
Good call :)

PR #1390, opened 2015-03-09
Add deploy:revisions_log

elia:
Show revisions log of the last 10 deploys.

Did a quick GitHub edit to start the discussion, will add specs and tune stuff if necessary.

cc @masterkain

### DIFF
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -177,6 +177,15 @@ namespace :deploy do
     end
   end
 
+  desc 'Show revisions log of the last 10 deploys'
+  task :revisions_log do
+    on release_roles(:all) do
+      within releases_path do
+        execute %{tail -n10 "#{revision_log}}
+      end
+    end
+  end
+
   desc 'Revert to previous release timestamp'
   task :revert_release => :rollback_release_path do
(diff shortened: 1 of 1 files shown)


leehambley:
Do we really need an internal task for `cat`ing a file on the server?

elia:
It's a `cat` (`tail`) on a server of a specific file in a specific path on a remote server to which you connect on some port with an identity file etc. …you got the idea ;)

I need it regularly, especially when pushing branches to staging servers. 
Doesn't seem out of place to me as it fits nicely under the definition of automation.
Also writing that file is part of the default deployment procedure and the filename is "owned" by capistrano itself.

That's all I got :)

leehambley:
Closing because it doesn't meet the contribution guidelines.

elia:
Does this mean that when will meet the contribution guidelines it will be merged?

I'm asking just to be sure this makes sense before trying to find the time to polish the PR.

leehambley:
It seems like it'll be a lot of work to add tests and everything for this, and I don't just know if it's worth it, it's only a one-liner, plus the boiler plate code, don't feel obliged to work on the PR, but if you can get it tested (which I suspect will be difficult) I'd be happy enough to merge it.

PR #1394, opened 2015-03-13
Don't do the work to lookup role names for substitution unless we have substitution.

znep:


leehambley:
Thanks for the contribution @znep can I point you at the CONTRIBUTING guidelines, and ask that you try and add a quick test case, and credit yourself in the CHANGELOG?

PR #1397, opened 2015-03-18
Bind post_task lazily in TaskEnhancements#after.

jcoglan:
Consider you have two files:

    # a.rake

    task :a do
    end
    after :a, :b

    # b.rake

    task :b do
    end

If `a.rake` is loaded before `b.rake`, it will fail, complaining that
"Don't know how to build task b". This is because TaskEnhancements#after
tries to look up task `b` before it's been defined.

However, the task is not actually needed until execution gets inside the
block that's passed to Rake::Task#enhance, at which point all tasks
should have been loaded and it's legitimate to complain about a task not
being recognised.

Havign this change allows app authors to write something like

    Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }

in their Capfile and not explicitly define load order. This cannot be
solved by calling `require` at the top of `a.rake` because `require`
throws a `LoadError` on `.rake` files. It also cannot be solved by
calling `import` at the top of `a.rake`, because Rake's `import` does
not load the referenced file immediately; it puts it in a queue and
continues processing the current file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 * Minor changes
   * Fix filtering behaviour when using literal hostnames in on() block (@townsen)
   * Added options to set username and password when using Subversion as SCM (@dsthode)
+  * Allow after() to refer to tasks that have not been loaded yet (@jcoglan)
 
 ## `3.4.0`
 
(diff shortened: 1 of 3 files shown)


leehambley:
Please see the contributing guidelines at https://github.com/capistrano/capistrano/blob/master/CONTRIBUTING.md otherwise looks fine (i.e just check the CHANGELOG, and credit yourself, etc)

jcoglan:
I've added a test for the issue I'm fixing, added myself to the changelog, and rebased.

leehambley:
Thanks @jcoglan !

PR #1423, opened 2015-04-24
Return longer SHA1 hash when fetching current revision

sds:
The mechanism Capistrano used to get the current commit hash of a
repository only returned the first 7 characters of the hash (thanks to the
`--abbrev-commit` flag).

While usually more convenient, the problem with a 7 character
abbreviation is that by the time a repository reaches 10,000 commits, it
becomes quite likely that you'll have a collision in the first 7
characters of at least two commits (~17% chance).

A collision is problematic if you intend to use these hashes to generate
links in emails or chat messages notifying your team of a deploy.

Increasing the minimum number of characters to 12 keeps the hash short
but virtually eliminates the possibility of a collision for all
practical purposes. (12 is the number used by the Linux kernel's git repo)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
   * Fix filtering behaviour when using literal hostnames in on() block (@townsen)
   * Added options to set username and password when using Subversion as SCM (@dsthode)
   * Allow after() to refer to tasks that have not been loaded yet (@jcoglan)
+  * Return first 12 characters (instead of 7) of SHA1 hash when determining current git revision (@sds)
 
 ## `3.4.0`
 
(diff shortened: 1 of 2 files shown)


z5h:
Can we just not abbrev-commit at all?
In short, if there is the possibility of any service/tool failing without a full SHA, then any gain of saving a few characters is moot.

Please read my comment here:
https://github.com/capistrano/capistrano/pull/1469

leehambley:
> While usually more convenient, the problem with a 7 character
abbreviation is that by the time a repository reaches 10,000 commits, it
becomes quite likely that you'll have a collision in the first 7
characters of at least two commits (~17% chance).

Good point, thanks for the patch, I've merged it now. Surprising that we didn't have a test for the abbreviated length of the commit hash. Thanks.

z5h:
There is still no answer for what important resource is saved by using a short SHA besides a few bytes of disk space. And why that is a priority over simply using a full SHA and ensuring with 100% certainty that SHA length will never be an issue again.

PR #1429, opened 2015-05-08
Allow use "all" as string for server filtering

theist:
Hi! 

Sorry I called it a bug but I do not know if it was a feature. 

Using capistrano I ran into a extrange behaviour, defined roles can be user as symbol or string for filtering, but special role all only could be specified as symbol

See relevant pry session at https://gist.github.com/theist/9ebf13d8aaee8a7ec01e

So wrote a test to match the behaviour I wanted and modify the Filter class code to pass it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
   * Added options to set username and password when using Subversion as SCM (@dsthode)
   * Allow after() to refer to tasks that have not been loaded yet (@jcoglan)
   * Ensure scm fetch_revision methods strip trailing whitespace (@mattbrictson)
+  * Allow use "all" as string for server filtering (@theist)
 
 ## `3.4.0`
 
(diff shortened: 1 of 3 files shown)


leehambley:
I believe that was actually by design, but it does rather look like it was confusing. Could you cleanup your commits based on the CONTRIBUTING guidelines (I see you touched the changelog) but nice commit messages are important, maybe rebase them into one commit, and include the contents of the gist in the commit message body, and use "Allow use "all" as string in additional to a symbol for server filtering"

theist:
Sorry, I've checked your contibuting gidelines but seems I didn't understood them completly.

I've rebased my branch (or, at least I hope I did it, if I did not, blame the google god)

leehambley:
> Sorry, I've checked your contibuting gidelines but seems I didn't understood them completly.

It's alright, it's more like a book!

> I've rebased my branch (or, at least I hope I did it, if I did not, blame the google god)

Rebase looks good, it's scary the first few times you use it. For what it's worth even if your base and screw something up the *commits* are never gone, they're just out of your history, all the commits stick around, and you can look them up, you can always good "git recover lost commits" or something. Git garbage collects" itself every ~30 days or something, but in that window, you can *usually* always recover from a rebase implosion!

PR #1432, opened 2015-05-16
Change host and role filtering configuration key

ynojima:
According to Capistrano document (http://capistranorb.com/documentation/advanced-features/role-filtering/) , filtering config takes "roles" and "hosts", but current implementation expects "role" and "host".

This pull request changes host filtering key from "host" to "hosts", and role filtering key from "role" to "roles".

Note: This breaks backward compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 
 * Removed the post-install message (@Kriechi)
 
+* Breaking Changes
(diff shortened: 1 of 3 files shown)


leehambley:
Don't you think it might be better to fix the documentation?

ynojima:
Yes, for backward compatibility, fixing the documentation would be better, but for unity among environment variable name, command line option, and configuration key, changing the key name would be better.
I don't have strong idea. If you value backward compatibility than unity, please close this pull request.

leehambley:
@townsen fixed this in 368ee4fc :) closing

PR #1435, opened 2015-05-26
Revert "Strip trailing whitespace for scm fetch_revision"

robd:
This reverts the stripping added in commit ad8682d. 

The original change to SSHKit which made it necessary to strip in capistrano was reverted in https://github.com/capistrano/sshkit/pull/249, so we no longer need to strip, at least for the moment.

If we update the SSHKit default capture behaviour in some future SSHKit release, we will need to revisit this, but for the moment I think it's best not to confuse things by implying stripping is needed and having tests which incorrectly show SSHKit as not stripping. Therefore, I kept the tests which were added, but updated the behaviour to match what SSHKit actually does.

@mattbrictson Since this reverts some of your original commit :(, I'd like to check whether this is OK with you.


mattbrictson:
@robd OK with me! :+1:

PR #1445, opened 2015-06-18
Ensure deploy:log_revision obeys `within`

theozaurus:
As the execute statement is not split by command and arguments the `within` block is ignored. Switching to `execute :cmd, "args"` solves this problem.


theozaurus:
Thanks!

PR #1447, opened 2015-06-24
`append` DSL method for pushing values like `linked_dirs`

kirs:
Before:
`set :linked_dirs, fetch(:linked_dirs, []) << 'vendor/bundle'`

After:
`append :linked_dirs, 'vendor/bundle'`

@leehambley @Kriechi

### DIFF
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -38,6 +38,10 @@ module Capistrano
       set(key, value, &block) unless config.has_key? key
     end
 
+    def append(key, value)
+      set(key, Array(fetch(key)) << value)
+    end
+
     def delete(key)
       config.delete(key)
     end
@@ -167,7 +171,7 @@ module Capistrano
       unless value.nil? or block.nil?
         raise Capistrano::ValidationError.new("Value and block both passed to Configuration#set")
       end
-      
+
       return unless validators.has_key? key
 
       validators[key].each do |validator|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 
+* `append` DSL method for pushing values like `linked_dirs`
 * Added support for git shallow clone
 * Remove 'vendor/bundle' from default :linked_dirs (@ojab)
 * Removed the post-install message (@Kriechi)
--- a/lib/capistrano/dsl/env.rb
+++ b/lib/capistrano/dsl/env.rb
@@ -6,7 +6,7 @@ module Capistrano
       extend Forwardable
       def_delegators :env,
                      :configure_backend, :fetch, :set, :set_if_empty, :delete,
-                     :ask, :role, :server, :primary, :validate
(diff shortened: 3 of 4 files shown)


leehambley:
can we rename it `set_append` or something? I like that set/fetch is symetrical, this seems like syntax sugar, which we always dented to avoid in favour of clarity.

kirs:
`set_append` sounds a bit confusing as a method name, because it contains two verbs. What about `push`? Or do you mean that you want to prefix the new method with `set_` anyway?

leehambley:
If it sets a variable, it should be called set!.

Maybe I'm tainted from too much Go, where append() is used like:

someVar = append(someVar, newVar)

lib/capistrano/configuration.rb
```diff
@@ -27,6 +27,10 @@ def set(key, value)
       config[key] = value
     end
 
+    def append(key, value)
+      set(key, fetch(key, []) << value)
```
Kriechi: Should we check that `fetch(key)` actually returns an array before using the `<<` operator?
Something other than an array might behave in different/unexpected ways...

leehambley: > Should we check that fetch(key) actually returns an array before using the << operator?

We could always just call `Array()` on it?

leehambley: @Kriechi ping I have some time for Capistrano releases today, if you can update #1447 I can get this out in 3.5.0 today.

Kriechi: that's actually a PR from @kirs - but I can do it as well...

leehambley: Oopse my mistake :) @kirs do you have time? (sorry I spent the time I had this morning clearing and merging the PR backlog, otherwise I'd take this)

kirs:
@leehambley done!

Kriechi:
@kirs could you rebase to current master as well? github won't merge it...

kirs:
@Kriechi done

leehambley:
@kirs are you seeing the same test failures as travis?

kirs:
@leehambley ooops I made a bad rebase, should be OK now

leehambley:
Thanks @kirs :)

mattbrictson:
Hey, this is a neat idea! @kirs can you fix the merge conflicts and then maybe we can get this merged?

kirs:
@mattbrictson done, thanks for update.

mattbrictson:
Oops, you need to also include the `append` method in `Capistrano::DSL::Env`, otherwise it is not accessible as part of the DSL.

Like this:

```ruby
# capistrano/dsl/env.rb
def append(key, value)
  env.append(key, value)
end
```

kirs:
@mattbrictson added.
Maybe we could use [delegation](http://ruby-doc.org/stdlib-2.0.0/libdoc/forwardable/rdoc/Forwardable.html#method-i-delegate) in future.

mattbrictson:
> Maybe we could use delegation in future.

Yes, in fact I already did this in https://github.com/capistrano/capistrano/commit/8c379b9312a143be9d4b9599b3fafd26f8a06929. If you want to rebase on top of `master` and follow my lead on that, please go for it. Or I can just merge this now and we can add the delegation later.

Issue #1451, opened 2015-07-01
Proposal: raise useful error message when capistrano-* gem is required incorrectly


mattbrictson:
One point of confusion with capistrano is the loading process for plugins. Many capistrano-* gems plug into capistrano by enhancing the `load:defaults` task. If this enhancement is done too late in capistrano's boot process, `load:defaults` has already been invoked. In this case the gem malfunctions because its setup process has been skipped.

Essentially the gem has to be required in the `Capfile`:

```ruby
require "capistrano/setup"
# plugins must be loaded here, after setup
```

If the user instead puts the require statement in `deploy.rb`, the plugin breaks, often in mysterious ways. See https://github.com/capistrano/rails/issues/37.

Now clearly this is a problem of not reading the documentation. But perhaps capistrano can make this mistake less likely or show a helpful error message in this scenario?

Here's an idea: in `capistrano/setup.rb`, alter the `load:defaults` task after invoking it, such that further attempts to enhance it will fail. Here is a very basic way to do it:

```ruby
invoke 'load:defaults'
Rake.application["load:defaults"].freeze
```

Now if a plugin loads out of order (i.e. in `deploy.rb`), it will immediately fail with `can't modify frozen Rake::Task`.

We could take this idea one step further and use `extend` to override `Rake::Task#enhance` to provide a more useful error message.

What do you think? Is this a worthwhile effort to make capistrano more forgiving/easier to troubleshoot?

leehambley:
I think that's a great suggestion Matt, but I'd be concerned that it's a half-measure, depending on where plugins actually hook in.  Another step in the right direction, that is for sure. Your idea of `Rake::Task#enhance` to maybe check if the object is `frozen?` and raise an error message along the lines of "looks like your plugin is late to the party bro" might be worthwhile too.

Could you take a shot at a patch ?

mattbrictson:
@leehambley I'll take a shot at a patch and open a separate PR.

Seeing that this is a "half-measure" (I agree), do you have any thoughts on what a more comprehensive solution would be? Perhaps a more formal plugin API? Right now plugin authoring is a bit mysterious; even enhancing `load:defaults` is more of an undocumented trick than an official policy.

mattbrictson:
Closed via #1452.

PR #1452, opened 2015-07-02
Freeze load:defaults task after invoking it so further enhancements fail

mattbrictson:
This causes a warning to be printed if a capistrano plugin is loaded in `deploy.rb` instead of the proper `Capfile` location. Output in this scenario looks like this:

```
$ cap production deploy:check
WARNING: load:defaults has already been invoked and can no longer be modified. C
heck that you haven't loaded a Capistrano plugin in deploy.rb by mistake. Plugin
s must be loaded in the Capfile to initialize properly.
(Backtrace restricted to imported tasks)
cap aborted!
can't modify frozen #<Class:#<Rake::Task:0x007fc1ef472d28>>

Tasks: TOP => production
(See full trace by running task with --trace)
```

Addresses #1451.

@leehambley This PR is incomplete, because I still have to figure out how to write an appropriate test. In the meantime, does my approach look good to you so far?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
   * Allow after() to refer to tasks that have not been loaded yet (@jcoglan)
   * Ensure scm fetch_revision methods strip trailing whitespace (@mattbrictson)
   * Allow use "all" as string for server filtering (@theist)
+  * Print a warning and abort if "load:defaults" is erroneously invoked after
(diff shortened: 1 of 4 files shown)


leehambley:
Hey @mattbrictson that looks super clean, I would have wondered if there was something other than `Rake::Task#extend()` that we needed to hook into, but it [doesn't look like it](http://ruby-doc.org/stdlib-2.0.0/libdoc/rake/rdoc/Rake/Task.html#method-i-reenable).

mattbrictson:
@leehambley Thanks for checking this out. I've now added tests, a CHANGELOG entry, and rebased this PR down to one commit. Ready to merge?

PR #1465, opened 2015-07-16
fix 'filter' variable so it works as documented

### LINKED ISSUE #1457: set :filter, :roles => %w{web} not working
Hi, in referencing this page: http://capistranorb.com/documentation/advanced-features/role-filtering/
I would assume putting `set :filter, :roles => %w{web}` in my deploy.rb file it would work (by only deploying to the servers listed in the web role. However, it doesn't work.

To get it to work I have to use either command line: `ROLES=web cap production deploy` or `cap --roles=web production deploy`. 

I'm using 3.4.0

townsen:
The variable ':filter' is documented as requiring the keys 'hosts' and
'roles' to set Host and Role filters. In actual fact it uses 'host' and
'role'. This commit adds support and tests for both.

Fixes #1457

### DIFF
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -100,6 +100,8 @@ module Capistrano
       @filters << Filter.new(:role, ENV['ROLES']) if ENV['ROLES']
       @filters << Filter.new(:host, ENV['HOSTS']) if ENV['HOSTS']
       fh = fetch_for(:filter,{}) || {}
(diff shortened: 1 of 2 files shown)


leehambley:
Excellent work @townsen keeping both, makes absolute sense. I was afraid of fixing this one way or another. This is important enough to warrant pushing out a release today, I've a tonne of other things to do, but I hope to get this, and SSHKit out today (we need a fix over there, because of changes upstream to Net::SSH, anyway)

Issue #1196, opened 2014-11-24
set does not accept passing a block the easy way


gucki:
In capistrano 2 a block could be passed to `set`. Capistrano 3 always expects two parameters and passing a block like this is no longer possible. 

Does not work in capistrano 3 (but it should work!):

    set :rvm_ruby_version do
      within release_path do
        capture(:rvm, "current")
      end
    end

Works in capistrano 3 (but looks like an ugly hack):

    set :rvm_ruby_version, proc{
      within release_path do
        capture(:rvm, "current")
      end
    }

Why has this support been removed? Anything against adding it again?

xavierholt:
Old thread, but I got very confused this afternoon when trying to pass a delayed-execution block to the set functions.  I agree with @gucki that this doesn't match Ruby conventions / violates the principle of least surprise.

I don't think it needs to be a breaking change, though.  It should be possible to support both methods simultaneously.  Proposed patch: https://github.com/capistrano/capistrano/pull/1478

gucki:
Thansk! :+1:

PR #1478, opened 2015-08-18
Issue 1196: Let Configuration#set take a block.

xavierholt:
My proposed fix for https://github.com/capistrano/capistrano/issues/1196

Allows the behavior of the set functions to match Ruby conventions, but shouldn't interfere with any apps that are using the current method of passing in blocks for delayed execution.


leehambley:
Thanks, nice work :)

PR #1479, opened 2015-08-18
Refactor Filter to use strategies instead of case statements

cshaffer:
* Reduces complexity of Filter#initialize and Filter#filter
* Consolidate regex matching to a single mixin for host and role
filtering


leehambley:
Thanks @cshaffer :)

Issue #1400, opened 2015-03-19
Host comparison should also check for ports.


ShogunPanda:
Hi all!

Currently Capistrano does this:

```ruby
class Capistrano::Configuration::Server
  def matches?(other)
    hostname == other.hostname
  end
end
```

which only check against the hostname.
Should we include the port as well? Otherwise it's impossible to differentiate the server basing the full address.

```ruby
class Capistrano::Configuration::Server
  def matches?(other)
    hostname == other.hostname && port == other.port
  end
end
```

mattbrictson:
This seems like a very rare scenario that we don't need to support. As a work around, you could give each host+port combination a unique alias using SSH config `Host`s.

mattbrictson:
Re-opening as it seems multiple people have reported this.

ShogunPanda:
I thing you addressed this. Closing this one.

PR #1490, opened 2015-09-10
fix match server: different ports

maliqq:
Same hostname might be used as a gateway to different servers.

```ruby
set :user, `whoami`.strip
# first server
server "192.168.1.100:1234", roles: %i{web}
# second server, same hostname
server "192.168.1.100:1235", roles: %i{redis}, user: "redisuser"
```
In this case, we have 1st server with user "redisuser", unless we explicitly define :user for server 1

### DIFF
--- a/lib/capistrano/configuration/server.rb
+++ b/lib/capistrano/configuration/server.rb
@@ -62,7 +62,7 @@ module Capistrano
       end
 
       def matches?(other)
-        hostname == other.hostname
+        hostname == other.hostname && port == other.port
       end
 
       private
--- a/lib/capistrano/configuration/servers.rb
+++ b/lib/capistrano/configuration/servers.rb
@@ -11,7 +11,6 @@ module Capistrano
         new_host = Server[host]
         if (server = servers.find { |s| s.matches? new_host })
           server.user = new_host.user if new_host.user
-          server.port = new_host.port if new_host.port
           server.with(properties)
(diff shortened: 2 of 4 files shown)


leehambley:
I don't think you "fixed" anything, can you take a look at the test failures please.

mattbrictson:
@leehambley Perhaps we should merge this one? Provided the OP rebases against latest `master` so that we can sure the tests are still OK. (The only Travis failure currently is 1.9.3 which is no longer relevant.)

This would fix #1400, I believe.

leehambley:
I don't rely on this behaviour at least, and we need to decide for one way or another.

NoICE:
In which version is this merged in? I updated to 3.4.0 recently and got into this again. I think I did run into this when upgrading to capistrano to 3.x from 2.x (in 2.x you could have multiple roles on servers with same name and different ports, no problem. I think many users need this, because of the concept of SSH gateway.) I used 3.2.x and it did work with that version, but on 3.4 it doesn't. So I'd like to know in which version is fix for this merged in...

mattbrictson:
@NoICE click on the merge commit

screen shot 2017-05-20 at 10 48 54 am

…and you'll see that it was included in v3.5.0:

screen shot 2017-05-20 at 10 49 03 am

NoICE:
Thanks :-)

PR #1492, opened 2015-09-16
Fixes #1221: Remove static path definition for git ssh wrapper

### LINKED ISSUE #1221: Dynamically set path to ssh executable
The path to ssh executable is hardcoded in `lib/capistrano/tasks/git.rake`[line 18](https://github.com/capistrano/capistrano/blob/master/lib/capistrano/tasks/git.rake#L18). In my case (a shared hosting webspace) the ssh executable is found at `/usr/local/bin/ssh`.
Before writing the git-wrapper script, this should be set dynamically with something like `whois ssh`.

As I am new to ruby, I didn't know exactly on which position this should be set. With a little help, I can do a pull request.

What do you think about my thoughts?

cheers
Jousch

signe:
Used bare 'ssh' rather than '$(which ssh)' because /bin/sh could be something other than bash (e.g. csh or bourne sh) on platforms such as Solaris.

Since $(which ssh) (or `which ssh`) is simply using the path, it serves no specific purpose in the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
     instead of `Capfile`. (@mattbrictson)
   * Return first 12 characters (instead of 7) of SHA1 hash when determining current git revision (@sds)
   * Clean up rubocop lint warnings (@cshaffer)
+  * Remove the static path to ssh in the git ssh wrapper to allow for greater
+    platform flexibility
 
 ## `3.4.0`
 
--- a/lib/capistrano/tasks/git.rake
+++ b/lib/capistrano/tasks/git.rake
@@ -15,7 +15,7 @@ namespace :git do
   task :wrapper do
     on release_roles :all do
       execute :mkdir, "-p", "#{fetch(:tmp_dir)}/#{fetch(:application)}/"
-      upload! StringIO.new("#!/bin/sh -e\nexec /usr/bin/ssh -o PasswordAuthentication=no -o StrictHostKeyChecking=no \"$@\"\n"), "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
+      upload! StringIO.new("#!/bin/sh -e\nexec /usr/bin/env ssh -o PasswordAuthentication=no -o StrictHostKeyChecking=no \"$@\"\n"), "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
       execute :chmod, "+x", "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
     end
   end


leehambley:
> Since $(which ssh) (or which ssh) is simply using the path, it serves no specific purpose in the code.

On the contrary, it's clear that it doesn't work my magic, and that people can look up in the `man` what `/usr/bin/env` does, and it's greppable, etc.

Changing this for everyone is such a probably huge breaking change, I can't imagine it passing review.

That said, if this is a problem for you, why not replace the SSHKit command map with a default that suits you:

``` ruby 
SSHKit.config.command_map = Hash.new do |hash, command|
  hash[command] = "/usr/local/rbenv/shims/#{command}"
end
```

see https://github.com/capistrano/sshkit#the-command-map

leehambley:
I'd consider replacing `/usr/bin/ssh` with `/usr/bin/env ssh` to make it clear that we defer to whatever semblance of a PATH there may be under this highly restrictive shell

signe:
Can you elaborate on why you think that this would be a breaking change?  I don't see that logic.  `/usr/bin` is part of the default path on every system I've used in the last 23 years, so assuming that there isn't another ssh binary that is higher in the path, it would still be the one that's used.  If there is one higher, then that's what the sysadmin would prefer was used anyway.

I can see using `env`, so I have updated the patch to use that.

I don't understand the SSHKit command-map referral.  This is a hardcoded string in a file that's executed on the remote system, not something that's being executed by the system running `cap`, and not dynamically created.

Currently overriding the path in `git-ssh.sh` requires this, which is entirely unhelpful and un-decipherable to most people.  (I had to find it referenced on StackOverflow, when I needed it.)

```
Rake::Task["deploy:check"].clear_actions

namespace :deploy do
    task check: :'git:wrapper'  do
        on release_roles :all do
            execute :mkdir, "-p", "#{fetch(:tmp_dir)}/#{fetch(:application)}/"
            upload! StringIO.new("#!/bin/sh -e\nexec /usr/local/bin/ssh -o PasswordAuthentication=no -o StrictHostKeyChecking=no \"$@\"\n"), "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
            execute :chmod, "+x", "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
        end
    end
end
```

mattbrictson:
This PR changes `/usr/bin/ssh` to `/usr/bin/env ssh`, which seems pretty benign to me. This is something that can't be done with the command-map, because the command in question is inside the `git-ssh.sh` script that gets uploaded.

So, 👍 from me.

ferdynator:
:+1:  Current way, described by @signe is pretty hard and definitely not straight forward.

will-in-wi:
I think this got lost in the shuffle. If it could be rebased on latest Capistrano, it sounds like we have consensus. This makes sense to me too. The biggest change is that `lib/capistrano/tasks/git.rake` has moved to `lib/capistrano/scm/tasks/git.rake`.

@signe, are you interested in doing this? Otherwise, I will.

PR #1494, opened 2015-09-21
Ignores Capistrano config files by default when deploying

cickes:
Capistrano config & task files were ending up on production servers,
which was less than ideal.

Base Capistrano files are now automatically ignored when deploying
via .gitattributes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+Capfile           export-ignore
+config/deploy.rb  export-ignore
+config/deploy/    export-ignore
+lib/capistrano/   export-ignore
+.gitattributes    export-ignore
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
     instead of `Capfile`. (@mattbrictson)
   * Return first 12 characters (instead of 7) of SHA1 hash when determining current git revision (@sds)
   * Clean up rubocop lint warnings (@cshaffer)
+  * Added default .gitattributes to ignore Capistrano files when deploying (@cickes)  
 
 ## `3.4.0`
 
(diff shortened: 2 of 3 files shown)


leehambley:
I might be misunderstanding something, but does this solve the problem of skipping uploading *Capistrano* when it's part of your `vendor/bundle`, and checked in?

leehambley:
Or, does Git collect the values of *all* .gitattributes files in all directories in the hierarchy, so having a a `.gitattributes` in Capistrano means that when running it, it's picked up from `vendor/bundle` and then Capistrano is ignored?

It seems, like this fix - if I've understood the intent correctly is a 1/2 measure, that'll work for people who vendor and commit their dependencies, but not those who don't.

Can't we solve this by maybe making these entries in `.gitattributes` after prompting the user for some confirmation, so that their own repo has the necessary fixes

cickes:
This skips uploading/deploying 
```lib/tasks/Capistrano/*
config/deploy.rb
config/deploy/*
Capfile
.gitattributes
```
when they are added to git version control (not .gitignored).  The files are kept in version control but excluded from deployment.

cickes:
> Can't we solve this by maybe making these entries in .gitattributes after prompting the user for some confirmation, so that their own repo has the necessary fixes  
  
This uses .gitattributes to `export-ignore` the files that you want commited to version control but not deployed.  All this request does is automatically ignore the base files created when the `cap install` command is executed locally.

leehambley:
But how does Git pick up files in Capistrano, if you don't commit your vendor/bundle, or if you don't use local bundle paths?

mattbrictson:
This is a cool idea, and I didn't know about `.gitattributes`. It would definitely be a nice thing to add if we can integrate it nicely (if not, maybe document it).

However, this PR as it stands will not work. It adds `.gitattributes` to the Capistrano repository, not to the *application* repository that is actually being deployed. As a result, it won't have any effect on actual deployments. (Unless I'm completely missing something here; please correct me if I'm wrong.)

Closing this PR because implementing this feature will require a different approach, IMHO.

PR #1501, opened 2015-09-29
Ensures that the git-ssh script is readable. Fixes #1500

### LINKED ISSUE #1500: git-ssh Permission Denied
In cases where read permission is not granted to the git-ssh.sh  file by default, one will receive a permission denied error during a deploy that uses git. (Remember that executing a script requires both execute AND read permission).

spight:
git-ssh requires both read and write permissions, and no checks exist before this point to ensure that the given ssh user has read access on this file, so let's do it here.

I looked for other instances that could be related to this issue, but a recursive grep for ":chmod" revealed only this single instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
   * Ensure task invocation within after hooks is namespace aware (@thickpaddy)
   * Deduplicate list of linked directories
   * Allow dot in :application name (@marcovtwout)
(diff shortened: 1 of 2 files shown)


mattbrictson:
This PR addresses issue #1500.

mattbrictson:
@spight Can you rebase this against current master and add a CHANGELOG entry? Then I will be able to merge this PR. Thanks!

spight:
@mattbrictson How's it look now?

mattbrictson:
Looks good. Thanks!

PR #1507, opened 2015-10-19
Require SSHKit >= 1.7.1

mattbrictson:
To make Airbrussh the default formatter in the next version of Capistrano (fingers crossed), we will need the latest version of SSHKit (currently 1.7.1). This is because Airbrussh itself depends on recent changes to SSHKit internals.

Is this reasonable? Capistrano has long had a very relaxed requirement for SSHKit, so this is definitely a change in policy.

/cc @leehambley @robd


leehambley:
Haters gonna hate, but sure - it's reasonable. Just to reiterate I try and be as liberal as possible with Cap dependencies because there's very, very little reason for your *development* deps to conflict with *production* deps, in this case, I suppose the minority of people who's *applications* use SSHKit is small enough that they can afford to upgrade, or jump through a hoop or two.

PR #1508, opened 2015-10-19
Isolate Gemfile used for Ruby 1.9 in Travis

mattbrictson:
Bundler sometimes resolves net-ssh 3.x when Travis builds against 1.9.3, and fails because these versions are incompatible. Work around this by explicitly specifying net-ssh ~> 2.8 when building against 1.9.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,11 @@ language: ruby
 rvm:
   - 2.1.0
   - 2.0.0
-  - 1.9.3
   - rbx-2
+matrix:
+  include:
+    - rvm: 1.9.3
+      gemfile: ruby19.gemfile
 script: bundle exec rake spec
 install: bundle install --jobs=1
 cache: bundler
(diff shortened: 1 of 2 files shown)


leehambley:
Thanks, I actually would just quietly drop 1.9 support, and mention it in
the release notes. It's near obsolete.

mattbrictson:
@leehambley Would you consider a similar PR for SSHKit? Or drop 1.9 support there as well?

leehambley:
Drop it everywhere to be honest, we have compelling 1.9x, working releases
out, 1.9 was EOL'Ed by Ruby upstream, what nearly 18months ago? I just
didn't remove it from Travis.yml yet :-)

robd:
On the SSHKit side, I wonder if it would be worth releasing latest master SSHKit with support for 1.9, and then dropping 1.9 support in the next major release.

leehambley:
I could live with that, but I'd rathe just quietly drop it, I know it's not exactly a smooth move to drop support without prior warning, but I already have a million things here, and at Capistrano to test.

mattbrictson:
Closed in favor of #1515

Issue #1512, opened 2015-10-23
:application validation is too strict and breaks existing projects; can't be disabled


mattbrictson:
I have many Rails projects using capistrano that have hyphens in the application name. As far as I can remember, this has never caused problems. Yet, the validation rule introduced in #1192 now prohibits my app from being deployed, with this error:

```
Invalid value for :application detected!
```

Now I am in a bit of a bind. I can't disable this new validation. Nor can I easily change my `:application` value, as this would have many consequences throughout my deployment (my app directory on the server would have to be renamed, for starters).

My thoughts:

1. Can we relax the validation rule to allow hyphens?
2. Is there a way to make validations opt-in or opt-out? (I think a good a solution is to have validation on by default for newly-generated projects, and off by default for existing projects.)

This of course assumes #1511 is fixed as well.

/cc @leehambley @Kriechi @ain

Kriechi:
I think you can remove a validation with something likes this (untested):
```ruby
validators[:application] = nil
```

As for the issue of `-` characters in application names, I would :+1: to include them into the existing validator - because @mattbrictson is right that this should be allowed. Could you send a PR to fix that?

mattbrictson:
@Kriechi Thanks for the tip on removing the validation. This works, but it is pretty obscure:

```ruby
env.send(:validators).delete(:application)
```

What do you think of this proposed addition to the DSL?

```
disable_validation :application
```

Then we can also improve the validation error message to look like this:

```
The :application value "my-site" is invalid!
Use only letters, numbers, hyphens, and underscores. For example:
  set :application, 'matt_site'
To silence this check, use:
  disable_validation :application
```

> As for the issue of - characters in application names… Could you send a PR to fix that?

Yes, I'll send a PR.

mattbrictson:
Or perhaps the more concise (and similar to `skip_before_action`, etc. in Rails):

```ruby
skip_validation :application
```

Kriechi:
I think skipping the `:application` validation makes no sense (once the validator is fixed). This validation is in place exactly to prevent Capistrano errors caused by filesystem things (spaces in folder names, etc.)

Since this is currently the only validation enforced by Capistrano itself, I'm against skipping it - and against added such possibilities (for now).

mattbrictson:
I agree that it is probably overkill to spend a lot of time on this given there is just one validation. Given that #1513 is merged I will close this issue.

PR #1513, opened 2015-10-23
Relax :application validator to allow hyphens

mattbrictson:
Also improve the error message to explain what characters are allowed, and include the offending value.

Addresses #1512.

Note: I could not find any existing test for the `:application` validator, and I wasn't sure how to add it. However since this is a small regex and string change, perhaps it can be merged as-is.


Kriechi:
:+1:

PR #1516, opened 2015-10-28
Fix post task lookup in TaskEnhancements#after

thickpaddy:
Since the change to lazily bind the post task in TaskEnhancements#after,
invocation of the post task is no longer done within the context of the
namespace within which after hooks are defined. This results in runtime
errors when after hooks are defined within a namespace and the post task
name provided as an argument does not include the full namespace.

For example, with the following code, tasks :bar and :baz will be
created within the foo namespace, but invoking foo:bar will cause
RuntimeError, "Don't know how to build task 'baz'" (note: no namespace)

    namespace :foo do
      task :bar do
        ...
      end
      after :bar, :baz do
        ...
      end
    end

Fixed by providing the scope of the after when looking up the post task.


leehambley:
+1 for tests, please check contribution guidelines and credit yourself int he changelog, or confirm you don't want to be credited and I'll merge it.

thickpaddy:
Sorry bout that, have updated changelog :-)

PR #1517, opened 2015-10-28
Make git wrapper work better with multi-user deployment systems

thickpaddy:
The wrapper script is known to cause problems on multi-user deployment systems, because the script is uploaded to the same location within tmp_dir, resulting in permissions issues when multiple users attempt to write to the same file, which is (sensibly) only writeable by its owner.

Using the application name in the path goes some way to addressing this, but doesn't help when multiple users deploy the same application.

I've worked around this by updating the path so it should "just work" on most systems by using a different file name for each deployer/application/stage. This seems to work well, but it does look a little fugly.

I've also made the path to the git wrapper script configurable, to make it simpler for devs and sysadmins to customise the behaviour to suit their systems if necessary, and to allow the path be referenced in any custom cleanup tasks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
+  * Make path to git wrapper script configurable (@thickpaddy)
+
 ## `3.5.0`
 
 https://github.com/capistrano/capistrano/compare/v3.4.1...v3.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
   * Make path to git wrapper script configurable (@thickpaddy)
+  * Change git wrapper path to work better with multiple users (@thickpaddy)
 
 ## `3.5.0`
 
--- a/lib/capistrano/tasks/git.rake
+++ b/lib/capistrano/tasks/git.rake
@@ -4,7 +4,10 @@ namespace :git do
   end
 
   set :git_wrapper_path, lambda {
-    "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
+    # Try to avoid permissions issues when multiple users deploy the same app
+    # by using different file names in the same dir for each deployer and stage.
+    suffix = [:application, :stage, :local_user].map { |key| fetch(key).to_s }.join("-")
+    "#{fetch(:tmp_dir)}/git-ssh-#{suffix}.sh"
   }
 
   set :git_environmental_variables, lambda {
--- a/lib/capistrano/tasks/git.rake
+++ b/lib/capistrano/tasks/git.rake
@@ -3,6 +3,10 @@ namespace :git do
     @strategy ||= Capistrano::Git.new(self, fetch(:git_strategy, Capistrano::Git::DefaultStrategy))
   end
 
+  set :git_wrapper_path, lambda {
+    "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
+  }
+
   set :git_environmental_variables, lambda {
     {
       git_askpass: "/bin/echo",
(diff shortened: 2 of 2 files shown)


thickpaddy:
Damn, this isn't going to be sufficient if two people deploying an application have the same local user and host name, which is quite plausible with Mac's being assigned default host names like "MacBook-Pro".

Back to the drawing board :-(

Suggestions welcome :-)

thickpaddy:
Another option is to use a random filename and add a cleanup task to delete it after the deploy. Does this sound ok? It would keep things quite simple.

leehambley:
The better solution even yet would be to make the git wrapper a shell script which uses an exit trap to remove itself, but it places a higher burden on catching all the places that depend on having the wrapper available, and it makes it difficult for 3rd parties such as the git-submodules plugin to make use of the same.But anything other than a tmpfile with a trap is not sane. (we did the best we could)

thickpaddy:
In principle that sounds a neat idea, but it wouldn't it also mean a separate git wrapper tmpfile is created for each task that requires the git wrapper as a prerequisite. Maybe that's no bad thing, but it also sounds like it would require extensive testing, whereas a git wrapper per deploy shouldn't as it's not much of a deviation from what's there already.

What do you think about allowing the git wrapper path to be configurable, to make it simpler to customise when the default doesn't work well for some people?

I'll have another look at this when I get a chance, it would be nice if it just worked, without being over-complicated. Right now, we have different tmp dirs for each user, which works, but damn it's ugly.

leehambley:
> What do you think about allowing the git wrapper path to be configurable, to make it simpler to customise when the default doesn't work well for some people?

Yep, exactly, more than this doesn't really work in reality. We could make the value a proc/lambdad have it be evaluated on runtime.

thickpaddy:
Cool, well, that bit is done in the first commit :-)

I'll have another look at ensuring the default path works in most, or at least more, contexts soon.

thickpaddy:
Hi Lee,

I've simplified the default git wrapper path to just generate a file name based on the application, stage and local user. I know this isn't ideal, but it's an improvement that should work most of the time, and when it fails it will be easy to understand why and work around it. I decided not to generate a random filename for because the idea of adding a special after hook to cleanup the tmp files just makes me uncomfortable (too much of a dependency between the internals of git related tasks and the general flow of deployment tasks).

Mark

leehambley:
I was reading he old doff when you pushed, looks sane. It strikes me that
we can't test the set() precedence is still forceable from the user configs
(Capfile) but it looks like it should work!

thickpaddy:
Sorry about the multiple force pushes, I ran into some bundler deployment problems testing with the gem on the local filesystem so pushed to github to test. I should have created a wip branch for testing rather than update the pull request.

Testing manually with an existing app worked fine though (well, once I'd fixed the bug in the code to fetch the config vars, oops). Also tested overriding the git_wrapper_path in my config/deploy.rb and that was fine too.

thickpaddy:
Hi @leehambley,

Just wondering what your thoughts are on this. Are you happy with the changes to make the path to the git wrapper script configurable and make the default path less likely to cause problems with multi-user deployment systems? If you are happy enough with the changes, but have reservations about merging due to the lack of test coverage, I can have a look at addressing that. Let me know what you think.

Thanks

Mark

leehambley:
@thickpaddy git wrapper path looks already to be configurable, I checked in master, I think that's why this branch is conflicting ?

thickpaddy:
Sorry, I hadn't spotted that there is now a conflict on this branch, I'll sort that out asap.

I don't think the conflict is due to the line you referred to, which makes it possible to change the GIT_SSH environment variable, but doesn't affect the uploading of the git wrapper script, or the destination path for that script on the servers. The only use I can think of for this is someone who already creates their own wrapper script and wants to use that rather than the one uploaded by capistrano.

thickpaddy:
@leehambley conflicts sorted, just a changelog clash

ingk:
Hi @thickpaddy, are there any news about this pull request?

mattbrictson:
We need @thickpaddy to fix the merge conflicts, and then this is good to go IMHO.

thickpaddy:
Hey, sorry about the delay following up on this, have been on holiday and just started a new job.

I've just fixed the conflicts, though I'm not actually in a position to test this right at the moment. Any chance someone else could give it a quick go to verify I haven't done anything stupid?

If not, I'll get to it later in the week, I just need a little bit of time to set up a local vm to deploy to.

mattbrictson:
I tested a deploy and also ran the feature tests: all OK. Looks great. Thanks!

thickpaddy:
Nice one @mattbrictson, thanks :-)

mattbrictson:
@thickpaddy OK, perhaps I didn't do a good job of testing this after all. It seems to have broken `deploy`.

The problem is that the `:git_wrapper_path` is not being used by `:git_environmental_variables`. Can you open a PR with a fix?

thickpaddy:
@mattbrictson oops, sorry, will open another PR in a mo

chewi:
@thickpaddy, just wanted to say thanks for doing this! We ran into the issue recently.

PR #1521, opened 2015-10-31
Change local_user default to use environment vars instead of getlogin

### LINKED ISSUE #1520: local_user unreliable on Mac OS X
Capitrano's `local_user` uses ruby's `Etc.getlogin` under the hood. Unfortunately that's not reliable on at least Mac OS X, because it is a global setting that returns whatever the last call to setlogin(2) was for the current login shell (quote from `man 2 setlogin` on Mac OS X 10.11.2):

> The getlogin() routine returns the login name of the user associated with the current session, as previously set by setlogin().  The name is normally associated with a login shell at the time a session is created, and is inherited by all processes descended from the login shell.  (This is true even if some of those processes assume another user ID, for example when su(1) is used.)

This means that if some background process started from my login session calls setlogin, the next time getlogin is called it will return that value shared across all processes that descend from the login shell.

**TL;DR:** Etc.getlogin might return names like `_mbsetupuser` or `_spotlight` and cannot be trusted.

Instead we could be relying on either `ENV['USER']` and falling back to `Etc.getlogin` if it's not set – in fact ruby's getlogin uses ENV['USER'], if the OS does not implement the getlogin syscall. Or we could use `Etc.getpwuid.name` which looks up the username from /etc/passwd given the current processes' uid obtained via getuid(2).

I'm not sure if capistrano supports Windows, if it does, we could use ENV['USERNAME'] on it.

felixbuenemann:
The return value of Etc.getlogin is not reliable on at least BSD style systems, because it can be changed any time by processes using the setlogin syscall, even after su to another user.

Instead we rely on USER (BSD style), LOGNAME (SystemV style) or USERNAME (Windows) environment variables. On modern unix systems both USER and LOGNAME should be set, but it doesn't hurt to check for both to be on the safe side.

This also removes the dependency on the Etc module and fixes #1520.

I have removed the dsl spec for local_user, because it gave the wrong impression of testing the default behavior when in reality it only exercised the dsl's set method.

### DIFF
--- a/lib/capistrano/defaults.rb
+++ b/lib/capistrano/defaults.rb
@@ -21,4 +21,4 @@ set_if_empty :log_level, :debug
 
 set_if_empty :pty, false
 
-set_if_empty :local_user, -> { Etc.getlogin }
+set_if_empty :local_user, -> { ENV['USER'] || ENV['LOGNAME'] || ENV['USERNAME'] }
--- a/lib/capistrano/dsl.rb
+++ b/lib/capistrano/dsl.rb
@@ -1,4 +1,3 @@
-require 'etc'
 require 'capistrano/dsl/task_enhancements'
 require 'capistrano/dsl/paths'
 require 'capistrano/dsl/stages'
--- a/spec/integration/dsl_spec.rb
+++ b/spec/integration/dsl_spec.rb
@@ -467,36 +467,6 @@ describe Capistrano::DSL do
 
   end
 
-  describe 'local_user' do
-    before do
-      dsl.set :local_user, -> { Etc.getlogin }
-    end
-
-    describe 'fetching local_user' do
-      subject { dsl.local_user }
-
-      context 'where a local_user is not set' do
-        before do
-          Etc.expects(:getlogin).returns('login')
-        end
-
-        it 'returns the login name' do
-          expect(subject.to_s).to eq 'login'
-        end
-      end
-
-      context 'where a local_user is set' do
-        before do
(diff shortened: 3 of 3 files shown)


felixbuenemann:
@leehambley It would be great to have a spec that checks the defaults, but I'm not sure what's the best way to load the defaults into a spec. Any ideas?

mattbrictson:
Should a similar change be made to SSHKit as well? SSHKit uses `Etc.getpwuid.name` here:

https://github.com/capistrano/sshkit/blob/e06d6b8424979a6a3a2974f375490d2aa15c2ac4/lib/sshkit/host.rb#L29

And I remember we had to patch around it in Airbrussh to get tests passing on Windows:

https://github.com/mattbrictson/airbrussh/blob/2bbe4177d5e29b142068a9c8a85fdbe5f6412617/test/airbrussh/formatter_test.rb#L20

felixbuenemann:
Etc.getpwuid.name should be safe on unix, but won't work on Windows, because it relies on /etc/passwd, so at least there would have to be a fallback. I think we could use environment vars there as well.

leehambley:
@mattbrictson good catch, @felixbuenemann would you mind to make the same patch for SSHKit? I know it's not the domain where you're having problems but it'd be nice to get them both done.

felixbuenemann:
@leehambley Sure, I'll take care of it.

leehambley:
Thanks, I've to do a release last Friday already, and I'm late because my
son was sick all weekend, but maybe I can get it out tonight/tomorrow
night. Thanks x10

PR #1537, opened 2015-12-17
Fixes #1536: Allow dots in application

### LINKED ISSUE #1536: Allow dot in application_name
Is there a specific reason why `.` (dot) is not allowed in the `:application_name` variable? I would like to use it to match our vhost names.

marcovtwout:


leehambley:
:rocket:

PR #1538, opened 2015-12-21
Enabling the printing of config variables with certain command line p…

gerardo-navarro:
When developing new capistrano tasks, it's very helpful to inspect all defined config variables. For this purpose, the new branch contains a new parameter for the command line `--print-config-variables` or `-p` that displays the defined config variables before starting the deployment tasks.

This feature was really valuable to during development and debugging of custom capistrano tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,9 @@ $ bundle exec cap production deploy --prereqs
 
 # trace through task invocations
 $ bundle exec cap production deploy --trace
+
+# lists all config variable before deployment tasks
+$ bundle exec cap production deploy --print-config-variables
 ```
 
 ## Validation of variables
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 * Old versions of SSHKit (before 1.7.1) are no longer supported
 * Drop support for Ruby 1.9.3 (Capistrano may still work with 1.9.3, but it is
   no longer officially supported)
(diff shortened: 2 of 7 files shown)


leehambley:
Looks good, observations:

1. It's not clear that these variables might be changed by plugins.
2. Would it not also be useful to print the underlying SSH kit mechanisms, such as the command maps/mappings?
3. You should credit yourself in the `CHANGELOG.md` file.
4. It should be documented in README somewhere since this would be really useful for many people
5. Even I can't see how this might interact with plugins that set, or add variables.
6. This is a decent reason to discourage people from using `set()` and `fetch()` for what should be standard Ruby variables.

gerardo-navarro:
Hi Lee, thanks feedback. I try to incorporate some of your points.

Can you give me some more information on what underlying SSH kit mechanism variables you mean and how I can access them?

README.md
```diff
@@ -83,6 +83,9 @@ $ bundle exec cap production deploy --prereqs
 
 # trace through task invocations
 $ bundle exec cap production deploy --trace
+
+# lists all config variable before deployment tasks
+$ bundle exec cap production deploy --trace
```
mattbrictson: Shouldn't this be `--print-config-variables` ?

gerardo-navarro: Yes, you are totally right, my bad. Change it!

mattbrictson:
@leehambley Can you do the final check and merge for this one?

@gerardo-navarro Thanks for fixing that doc. To make it more likely that this PR will be merged, can you please rebase this down to a single commit?

leehambley:
Waiting for @gerardo-navarro

gerardo-navarro:
@leehambley : it is done!

I was on vacation without a notebook. Thanks for the help ...

PR #1541, opened 2015-12-26
Make Airbrussh the default formatter for Capistrano

mattbrictson:
To make Airbrussh the default formatter, this commit makes the following changes:

1. Adds the airbrussh gem as a dependency
2. Bumps the sshkit dependency to 1.8.0 (for the new `use_format` method)
3. Introduces a `:format_options` setting

Out of the box, Capistrano now defaults to the following:

```ruby
set :format, :airbrussh
```

It also requires the airbrussh gem at the appropriate point in the Capistrano boot process, which allows Airbrussh to register its `deploy:failed` hook.

We also change the default Airbrussh settings slightly, to acknowledge the fact that Airbrussh will now be the first impression new Capistrano users have of the tool.

Specifically:

* Enable command output (hiding the command output is tidy but probably will confuse new users who will not know what their deploy script is doing)
* Disable the "using airbrussh format" banner at startup

Finally, this commit adds an official mechanism for sending a hash of options to the formatter initializer. This lets users easily change Airbrussh's settings. E.g.:

```ruby
set :format, :airbrussh
set :format_options, color: false
```

**TODO**

These changes depend on changes to Airbrussh and SSHKit (hence the feature spec failures). So in order to merge this PR, first we need:

- [x] CHANGELOG
- [x] Merge https://github.com/mattbrictson/airbrussh/pull/70
- [x] Merge https://github.com/capistrano/sshkit/pull/308
- [x] Release new Airbrussh version
- [x] Release new SSHKit version
- [x] Update capistrano.gemspec to depend on these new releases


mattbrictson:
@leehambley What do you think about releasing a beta1 or rc1 of the next SSHKit version so that we can start testing this end-to-end?

leehambley:
Let's do it - would you like me to take care of it ? I have to check whether the CHANGELOG is upto date, I think it should be, but I can push an rc if you like.

leehambley:
Including this commit, I assume you mean?

mattbrictson:
No, I mean cut a release of SSHKit, not Capistrano. This PR is for Capistrano but I need a new SSHKit version because some of the changes in this PR depend on as-yet-unreleased APIs in SSHKit.

mattbrictson:
I don't have rubygems push rights for sshkit, but if you would like to grant those I will be happy to take care of the rc1 release.

leehambley:
I'll drop you a note when I've had time to add you mate, a bit hectic at
home tonight.

mattbrictson:
Whew! This PR has been a long time coming, but I believe it is finally done and ready to merge.

Note that the gemspec points to pre-release versions of Airbrussh and SSHKit; we will need to switch those to final versions before we actually release Capistrano 3.5.0.

```ruby
gem.add_dependency 'airbrussh', '>= 1.0.0.beta1'
gem.add_dependency 'sshkit', '>= 1.9.0.rc1'
```

@leehambley can you do the honors? :smile:

PR #1542, opened 2015-12-28
Fix broken deploy_failure.feature tests

mattbrictson:
It seems that enhancing a task using an `after` block will not have desired effect if the new task shares the same name (excluding namespace) as the task being enhanced. In other words, `after 'deploy:failed', :failed do ...` will not work (the block will never be executed). This behavior was recently introduced in 5fcf3da.

To fix the feature test using this code, I renamed the task to be a unique name. The test now works as intended.

Also, I noticed the task defined in `fail.rake` was sometimes not working as intended, because the `shared_path` did not exist on the VM's filesystem. This meant that `touch` was failing. Corrected by issuing `mkdir -p` first.

With these changes, all feature tests now pass.

### DIFF
--- a/spec/support/tasks/fail.rake
+++ b/spec/support/tasks/fail.rake
@@ -1,6 +1,7 @@
 set :fail, proc { fail }
 before 'deploy:starting', :fail do
   on roles :all do
+    execute :mkdir, '-p', shared_path
     execute :touch, shared_path.join('fail')
   end
   fetch(:fail)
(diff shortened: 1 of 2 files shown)


leehambley:
If I'm reading this correctly @mattbrictson this doesn't solve the problem for which we have another ticket, but rather works around it by making our tests more robust, and not shadowing another task name?

mattbrictson:
> If I'm reading this correctly @mattbrictson this doesn't solve the problem for which we have another ticket, but rather works around it by making our tests more robust, and not shadowing another task name?

This is correct. I wasn't sure what the resolution of the `after` issue would be, so I worked around it.

PR #1547, opened 2015-12-29
Normalize before/after task defn. and invocation

### LINKED ISSUE #1543: Using before/after with a fully-qualified (namespaced) task has unexpected behavior
Capistrano's `before` and `after` work with `namespace` blocks, thanks to 5fcf3da. Given this:

```ruby
namespace :deploy do
  after :updated, :foo do
    # ...
  end
end
```

Here everything works as expected:

1. a `deploy:foo` task is created
2. the existing `deploy:updated` task is enhanced
3. at runtime, `deploy:foo` is invoked after `deploy:updated`

However, doing the same thing by fully-qualifying the task name instead of using the `namespace` block leads to some very odd behavior. Consider:

```ruby
after "deploy:updated", :foo do
  # ...
end
```

In this case:

1. a `foo` task is created at the root level (as expected)
2.  the existing `deploy:updated` task is enhanced
3. at runtime, **`deploy:foo`** (huh?) is invoked after `deploy:updated`

When `deploy:updated` is invoked, Capistrano tries to find and invoke `deploy:foo`. The only reason the deploy still works is because Rake is smart enough to also search for the root `foo` when it fails to find `deploy:foo`.

This is confusing, and I think dangerous.

Consider this example (see #1542):

```ruby
after "deploy:failed", :failed do
  # ...
end
```

This task enhancement doesn't work! Even though a new root level `failed` task has been defined, Capistrano looks up `deploy:failed` at runtime instead. So essentially what has been accomplished is `deploy:failed` has been enhanced with itself! The root `failed` task is never invoked.

I think the desired (and intuitive) behavior is that the task being defined in the `after` block should be the same task that is invoked.

mattbrictson:
Capistrano and Rake both have ways of enhancing existing tasks with additional actions or prerequisites, but the behavior is subtly different.

In Capistrano, before/after implicitly create a new named task (when a block is supplied). When this feature is used, task creation and task invocation take place in two different runtime contexts, which leads to subtle inconsistencies with regards to namespaces.

Rake's solution is to *always* search the task's namespace when looking up a prerequisite. E.g. if we add `foo` as a prerequisite like this:

```ruby
Rake::Task['deploy:starting'].enhance([:foo])
```

Then during task invocation, Rake will look for `foo` in the `deploy` namespace. If it can't find `deploy:foo`, then it falls back to looking for `foo`. (I find this behavior non-intuitive, but that's how it works.)

This is at odds with how Capistrano's system behaves. If we try to accomplish the same thing using Capistrano `before` block syntax:

```ruby
before('deploy:starting', :foo) do
  # ...
end
```

This creates a root-level `foo` task, but then since Capistrano uses Rake's prerequisite system under the hood, at invocation time `deploy:foo` is invoked. **The task being declared and the lookup being performed don't match.**

This commit fixes this inconsistency. Now `foo` is created at the root level as it always has been, but the invocation behavior is changed: Capistrano now correctly invokes the root-level `foo` instead of searching the parent tasks's namespace.

Accomplishing this required some tricky code involving Rake's special `rake:` namespace to force the task lookup behavior we want, and definitely goes against the grain of Rake. However this seemed to be the best solution that makes the task declaration and invocation behavior consistent when `before` and `after` are used.

No existing tests broke as a result of this change, which speaks well for backwards compatibility. I added more tests to cover the edge cases that are impacted.

Fixes #1543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,22 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 * Drop support for Ruby 1.9.3 (Capistrano may still work with 1.9.3, but it is
   no longer officially supported)
 
+* `before` and `after` hooks have changed slightly:
+  * Tasks created with a `before` or `after` block within a `namespace` are now
+    executed properly. In prior Capistrano versions, namespaced `after` tasks
+    would fail to be found. (@thickpaddy)
+  * Tasks attached with `before` or `after` outside of a `namespace` are
+    now invoked as root-level tasks. In prior Capistrano versions, a `before`
+    task hooked onto a namespaced task (e.g. `deploy:starting`) would inherit
+    that task's namespace. Note that this behavior is different from Rake's
+    prerequisites feature, where the namespace of the original task is always
+    searched when looking up prerequisites. Your `before` hooks may break if
+    you relied on the Rake-style behavior. (@mattbrictson)
+  * `after` can now refer to tasks that have not been loaded yet (@jcoglan)
+
 * Minor changes
   * Fix filtering behaviour when using literal hostnames in on() block (@townsen)
   * Added options to set username and password when using Subversion as SCM (@dsthode)
-  * Allow after() to refer to tasks that have not been loaded yet (@jcoglan)
   * Ensure scm fetch_revision methods strip trailing whitespace (@mattbrictson)
     * Reverted - no longer needed due to [SSHKit PR249](https://github.com/capistrano/sshkit/pull/249) (@robd)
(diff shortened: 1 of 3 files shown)


mattbrictson:
Argh. I tested this with a Rails deploy and it breaks the capistrano-bundler plugin. The plugin incorrectly double-specifies the namespace like this:

```ruby
namespace :bundler do
  before "deploy:updated", "bundler:install"
end
```

Now that Capistrano more strictly follows namespacing rules, it tries to find the task `bundler:bundler:install`, which fails.

I guess there isn't a way to solve this without breaking existing plugins.

PR #1554, opened 2016-01-06
Continue if HOSTROLEFILTER= is set

eric:
I noticed in v2 that `HOSTROLEFILTER=` behaves differently than using `HOSTFILTER=`. This will make them both behave in the same way and skip tasks when there are no servers matched.

### DIFF
--- a/lib/capistrano/configuration/connections.rb
+++ b/lib/capistrano/configuration/connections.rb
@@ -153,7 +153,7 @@ module Capistrano
           servers = find_servers_for_task(task, options)
 
           if servers.empty?
-            if ENV['HOSTFILTER'] || task.options.merge(options)[:on_no_matching_servers] == :continue
+            if ENV['HOSTFILTER'] || ENV['HOSTROLEFILTER'] || task.options.merge(options)[:on_no_matching_servers] == :continue
               logger.info "skipping `#{task.fully_qualified_name}' because no servers matched"
             else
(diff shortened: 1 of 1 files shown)


leehambley:
Hi @eric nice to cross paths again. We don't really do support for v2 since none of the maintainers actually use it anymore.

At some point we wanted to stop accepting patches for v2, I thought we had been pretty transparent about it being EOL. I'm not actually sure if we should accept this patch and risk another release - or whether we can use this opportunity to ask what it would have taken to make it clear to you that we don't take code for v2 anymore? (we always always always made "one last exception" to that rule the last 5 times someone opened a PR)

eric:
I'm not really sure what the right solution is, but I haven't had the time to step back and evaluate what would be required to port all of my capistrano recipes to v3 (because v2 works for me and changing these things can often lead to subtle issues and bugs).

I would definitely appreciate if you continued to accept small fixes, but if you don't want to, I guess I can start running a fork that fixes any additional little issues I run into.

leehambley:
If you're stuck on v2, we could make you the man who decides if PRs work on v2?

eric:
Sounds good to me.

leehambley:
Great, so if you can prep a release PR with appropriate tags/etc, I'll pull it and build and release the gem (assuming it all works.) Thanks @eric

PR #1559, opened 2016-01-17
Rewrite of the Capistrano README

mattbrictson:
I was in a documentation writing mood, so I rewrote the README. :grin: 

While Capistrano does have official documentation on its website, I feel that we still need a comprehensive README. (In fact, the website instructs readers to read the README first.)

My goal here was to do a better job of explaining what exactly Capistrano is, why you should use it (there are many options out there for deployment automation, after all), and also importantly, what Capistrano does *not* do.

I also tried to make a more helpful "documentation and help" section, and I removed contribution and testing instructions that are better covered in the new CONTRIBUTION and DEVELOPMENT docs.

Finally, I removed the section about "Validation of variables", which seemed out of place in the README. I think this is more appropriate for the documentation website, since it is a pretty low-level feature.

I realize that the "features" and "gotchas" sections that I've written are quite wordy. If you feel it is better to move these to the documentation website in order to keep the README more concise, I can do that.


leehambley:
Nothing I would change. Vast improvement.

mattbrictson:
Thanks! I'll merge it in, then.

Issue #929, opened 2014-02-10
Do not load SCM tasks unless an SCM is defined


leehambley:
Related to #718 and a handful of other issues which relate to deploying without an SCM, unloading the SCM, deploying via copy, tar ball, BitTorrent or quantum entanglement.

I propose, based on the work done in #718 to only load the flow, that without `set(:scm, “something”)` we should not load an SCM at all.

I propose that we could solve this by doing runtime class resolution, and autoloading of files incase a class is undefined, which may mandate changes to the structure of the SCM code. In that case we would still load the `git` namespace from `git.rake` (and others) - but avoid, if possible inserting the tools, which we could do in an `include` or similar hook in `git.rb`

fazy:
Adding to the discussion of use cases for deployment without SCM, it's already been mentioned that a solution like Bamboo will create a package with the full, built application ready to unpack and run. For some users, this is preferable to pulling source code and building the application during the deployment process. For one thing, if you use tools like NPM (Node) or Composer (PHP), then you don't want to be dependent on so many third party resources (if I managed to get it all downloaded and working at build time, I would like to keep this for deploy).

I'm now looking at using Capistrano to deploy to a group of Amazon EC2 instances. The [cap-c2](https://github.com/forward3d/cap-ec2) project looks like it provides good support.

EC2 instances are ephemeral and elastic; new ones might be started at any time. My thinking is to have the build package available in S3, and when an instance starts for the first time it can fetch the latest build and run a local deploy script. But for already running instances, I would still like Capistrano to orchestrate this.

So I don't need scm:copy, I need either scm:none, or for the default scm:git to not get in the way.

Is this currently possible, or maybe the use case is too far removed from what Capistrano aims to do? I'm unlikely to be much help with creating a plugin as my entire Ruby knowledge is confined to what's visible in the setup guide of Capistrano. ;)

johnkharvey:
I found a workaround-- sort of.

Here's part of my config/deploy.rb:

```
# config valid only for Capistrano 3.2.1
# Capistrano 3.3.3 includes "statistics",  which is not desirable and could not be turned off yet
lock '3.2.1'

load 'config/my_functions.rb'

set :application, 'gitignore'
set :repo_url, 'https://github.com/github/gitignore.git'

########################################################
# Unfortunately, in capistrano 3.x, you need to specify
#   an SCM target, so these variables all impact that
########################################################
set :deploy_to, '/home/someuser/git/gitignore'
[...]

PR #1561, opened 2016-01-18
Add Capistrano Plugin API

mattbrictson:
This PR is based on my discussion with @leehambley [here](https://github.com/capistrano/capistrano/issues/929#issuecomment-170371959) and my proof-of-concept [here](https://github.com/mattbrictson/experimental-capistrano-git-scm).

I have refrained from any CHANGELOG entry because I don't think we should publicize the Plugin class too much until it has had some real-world use and the API is proven to be stable.

SSHKit doesn't yet support `SSHKit::Backend.current` (I need to open a PR for that), so in the meantime I've added it here as a monkey patch.

I believe everything is well unit-tested. (Please forgive me if my RSpec dialect is awkward; I usually write tests using Minitest. :sweat:)

### DIFF
--- a/lib/capistrano/ext/sshkit/backend/thread_local.rb
+++ b/lib/capistrano/ext/sshkit/backend/thread_local.rb
@@ -1,0 +1,25 @@
+require "sshkit/backends/abstract"
+
+# Monkey patch older versions of SSHKit to make the currently-executing Backend
+# available via a thread local value.
+#
+# TODO: Remove this code once capistrano.gemspec requires newer SSHKit version.
+
+unless SSHKit::Backend.respond_to?(:current)
+  module SSHKit
+    module Backend
+      def self.current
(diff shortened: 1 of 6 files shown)


mattbrictson:
Hmm. Looks like Travis is not happy. I'll look into the test failures.

mattbrictson:
Tests fixed. This is ready for review!

mattbrictson:
@leehambley I just realized that this PR introduced code with keyword arguments, which breaks Ruby 1.9.3. Of course we don't officially support 1.9.3, but this would be the first change (I think) that explicitly breaks it.

Are you OK with that? I can always change the code to use an options Hash instead.

leehambley:
> Are you OK with that? I can always change the code to use an options Hash instead.

I dare say that if we're releasing a new version with a bump in the number we can afford to warn of a breaking change.

PR #1562, opened 2016-01-18
Refactor display_error_message to simplify.

will-in-wi:
Noted that Code Climate found this method complicated, and I see why. Refactor to simplify.

### DIFF
--- a/lib/capistrano/application.rb
+++ b/lib/capistrano/application.rb
@@ -63,7 +63,7 @@ module Capistrano
 
     def display_error_message(ex)
       unless options.backtrace
-        Rake.application.options.suppress_backtrace_pattern = backtrace_pattern
+        Rake.application.options.suppress_backtrace_pattern = backtrace_pattern if backtrace_pattern
         trace '(Backtrace restricted to imported tasks)'
       end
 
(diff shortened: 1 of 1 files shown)


lib/capistrano/application.rb
```diff
@@ -63,13 +63,10 @@ def top_level_tasks
 
     def display_error_message(ex)
       unless options.backtrace
-        if (loc = Rake.application.find_rakefile_location)
-          whitelist = (@imported.dup << loc[0]).map{|f| File.absolute_path(f, loc[1])}
-          pattern = %r@^(?!#{whitelist.map{|p| Regexp.quote(p)}.join('|')})@
-          Rake.application.options.suppress_backtrace_pattern = pattern
-        end
-        trace "(Backtrace restricted to imported tasks)"
+        Rake.application.options.suppress_backtrace_pattern = backtrace_pattern
```
mattbrictson: This has the effect of setting `Rake.application.options.suppress_backtrace_pattern` to `nil` if `loc` is falsey, which wasn't the case before the refactor.

will-in-wi: Good catch, fixed.

mattbrictson:
Thanks, this seems like an overall improvement.

I'm not familiar with this part of the Capistrano code though, so I don't feel comfortable merging.

@leehambley can I get a second opinion on this one?

leehambley:
Also not familiar with it, but it looks sane. If we didn't break the tests, let's merge it :)

PR #1570, opened 2016-01-23
Fix overly-broad use of `raise_error` with no argument

mattbrictson:
Fixes these RSpec warnings:

> WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call.


leehambley:
Good housekeeping @mattbrictson :+1:

Issue #1125, opened 2014-08-29
Error "don't know how to build" when overriding cap SCM tasks


adeolaawoyemi:
Hi,

I'm trying to work around an svn authentication issue by overriding the `svn:update` task. My task runs fine but cap also runs the original tasks actions and according to the docs, when I use:

``` ruby
Rake::Task["svn:update"].clear_actions
```

I get the error:

``` shell
$ SVN_USERNAME=myuser SVN_PASSWORD=mypass cap qa deploy --trace
** Invoke qa (first_time)
** Execute qa
** Invoke load:defaults (first_time)
** Execute load:defaults
cap aborted!
Don't know how to build task 'svn:update'
...
```

How can I run my override task and clear_actions of predefined task without getting this error? Or is there another way to make the subversion command always use the defined username and password provided for ALL svn commands?

Thanks.

schnittchen:
This should be re-opened. Does not work as documented.

mattbrictson:
You're correct: clearing the default SCM tasks in particular doesn't work because the SCM tasks are not loaded and defined until *after* your deploy.rb is evaluated. Reopening.

mattbrictson:
This is a hack, and perhaps Capistrano should provide a better way of accomplishing this, but here is one solution.

Since the SCM task you want to override is not defined until later in the boot process, you need to delay the code that does the override. My solution is to put that code inside an `after` block that hooks into `deploy:starting`.

Then you can lookup the default task, clear it, and substitute your own implementation.

```ruby
after "deploy:starting", :override_scm do
  svn_update = Rake::Task["svn:update"]
  svn_update.clear_actions
  svn_update.enhance do
    puts "Running custom svn:update instead"
  end
end
```

leehambley:
That is unfortunate, good call @mattbrictson - I wonder if we can fix the load ordering, somehow - I would have lost hours to this.

mattbrictson:
> I wonder if we can fix the load ordering, somehow

@leehambley I'm not sure there is a way to fix this and guarantee backwards compatibility. We can't know for certain what SCM the user has specified until we evaluate `deploy.rb`, because that file might contain e.g. `set :scm, :svn`. But then the SCM isn't loaded yet, so task enhancements fail.

One solution would be to load *all* built-in SCMs up front. I don't see any downsides to having git, hg, and svn tasks defined; only the specified SCM is actually invoked during the deploy tasks.

[...]

Issue #1409, opened 2015-03-30
Add Git submodule support


paulschreiber:
Please add official submodule support. I spent a few hours this afternoon trying various submodule solutions found around the web.

Some of them do weird things, like `find .git` and `rm` or `rsync`. None of them seem to properly cache the submodule, which means it's getting cloned on every deploy, which is super inefficient.

dwradcliffe:
+1

leehambley:
This is largely blocked by https://github.com/capistrano/capistrano/issues/929 which would then allow us to have a `git` (naïve) and a git with submodule (or, one with some options) plugins. As it is, there's already a bit too much special-casing in the core.

tymat:
This needs to be added.  +1

mani95lisa:
+1

1st8:
I proposed a way to handle `:git_enable_submodules` in https://github.com/capistrano/capistrano/pull/1823 which works nicely for me so far.
Please let me know what you think.

aiyanbo:
+1

1st8:
Could you try https://github.com/ekho/capistrano-git-with-submodules ?
It works great for me.

leehambley:
Ekho's plugin is the way this was intended to be solved. The SCM plugin code that landed ~3.6-3.7 was intended to support just this use-case. Git submodules are not without their downsides and complexities, so we took care to make them pluggable so that people could pick their preferred method.

PR #1572, opened 2016-01-24
Deprecate `set :scm, ...` in favor of new SCM plugin system

mattbrictson:
- [x] CHANGELOG
- [x] Blog post/documentation explaining migration path for SCM plugin authors
- [x] Squash commits; add detailed commit message
- [x] Review @will-in-wi's feedback (see code comments below)

This PR implements the migration path from the old way of loading the SCM implementation, which is to say, from this:

```ruby
# deploy.rb
set :scm, :git
```

To this:

```ruby
# Capfile
require "capistrano/scm/git"
install_plugin Capistrano::SCM::Git
```

I do this in such a way that all existing Capistrano projects should continue to work unmodified. Internally the new SCM plugin will be used automatically and a deprecation warning will be printed. For example, if you are using `set :scm, :git`, you will see:

```
WARNING: `set :scm, :git` is deprecated.
         To ensure your project is compatible with future versions of
         Capistrano, remove the :scm setting and instead add these lines to
         your Capfile:

         require "capistrano/scm/git"
         install_plugin Capistrano::SCM::Git
```

If you are using a third-party SCM, like `:copy`, then you would be given a link to as-yet-unwritten blog post on the topic:

```
WARNING: Loading the copy SCM via `set :scm, :copy` is deprecated.
         To ensure this plugin will work with future versions of Capistrano,
         please upgrade it to a newer version. Refer to this post for more info:

         http://capistranorb.com/posts/...
```

All of this logic is pretty much self-contained in a single class: `SCMResolver`, which would go away once Capistrano 4.0 is released and we can drop backwards compatibility.

---

So, this PR is my way of asking:

* Does the proposed new way of loading SCMs make sense?
* Is the migration/deprecation path reasonable?

There is a lot more work to be done in terms of writing tests and adding documentation, so I'd like to make sure I have some buy-in from the Capistrano team before going further.

To recap, the motivation for this rewrite is to treat SCMs just like any other Capistrano plugin (i.e. loaded in the Capfile) so they can be customized, made optional, and moved to external gems; along the way fixing #929, #1125, and many others.

/cc @leehambley @robd


leehambley:
LGTM @mattbrictson - Surprising to see how much there is to do in the SCM resolver, actually!

mattbrictson:
> Surprising to see how much there is to do in the SCM resolver, actually!

Yeah, backwards-compatibility is hard!

One thing that I'm concerned about is that the new plugin style is a little more verbose. Instead of one line magically configuring the SCM, you now have to write two lines:

```ruby
require "capistrano/scm/git"
Capistrano::SCM::Git.new
```

So in some sense this is a step backwards. There is more to type, more chances to misspell something, etc.

On the other hand, there is less magic and it is just plain Ruby.

I considered a special `plugin` DSL method that would combine the steps, like this:

```ruby
plugin "scm/git"

# …which behind the scenes expands to
require "capistrano/scm/git"
Capistrano::SCM::Git.new
```

But so far I've decided against that as being too magical/proprietary.

What do you think?

leehambley:
> But so far I've decided against that as being too magical/proprietary.

Is there a middle ground where we can make your proposed two-liner the "I need a special case" way, and where we can still do something like `require "capistrano/scm/git/defaults"` for doing essentially this ?

mattbrictson:
> Is there a middle ground where we can make your proposed two-liner the "I need a special case" way, and where we can still do something like require "capistrano/scm/git/defaults"

One thing I want to avoid is an explosion of different requires or a "require-driven API" like many Capistrano plugins are currently. E.g. capistrano-rails is broken up into `rails`, `rails/assets`, and `rails/migrations`; and capistrano-bundler has `bundler`, `bundler/hooks`, and `bundler/tasks`. This scheme is not applied consistently across plugins, and I find using `require` to determine behavior to be kind of a blunt instrument.

But I do see the value of using it on some occasions, like how you can require `minitest` for general purpose or `minitest/autorun` to get the automatic `at_exit` behavior.

Hmm. I can't decide. :sweat_smile:

leehambley:
Right, minitest was what I had in mind, but the plugins with optional
behaviour are tricky. I literally never use the asset pipeline with Rails,
but always use migrations.... Hrmph.

mattbrictson:
Since it is a toss-up, I am going to err on the side of being more explicit. I'll leave it as two steps (`require`, `new`) for now.

leehambley:
Ok mate, cool.

mattbrictson:
@will-in-wi I was wondering if I could get your opinion on this PR? It is a notable change to how SCM is configured in Capistrano. It removes a lot of "magic" in favor of explicit Ruby, but that can also be a bad thing if it feels too cumbersome.

I feel nervous merging this with only me and @leehambley having reviewed it. Would love to get your input as well.

will-in-wi:
I'm happy to hear about the underlying code change. A pluggable SCM system makes a ton of sense. I was personally wondering just the other day if I could implement something that would download and extract a tarball, and it looks like this would be a lot easier in the future.

Per the user end of things, my initial thought on reading the description was that the new way is quite a bit more verbose. I don't mean in the form of two new lines, but rather that it seems to be breaking from how I would have expected Capistrano to work. I'm personally a fairly experienced Ruby developer, so seeing through the Capistrano DSL into the underlying Ruby isn't difficult for me. However, a lot of people use Capistrano who are not Ruby developers. This particular syntax shouldn't trip them up too much, but it is introducing more Ruby into the simpler Capistrano DSL that was formerly all you needed to know.

My other concern is that I feel like this syntax violates command/query separation to some extent. I usually don't expect to see a class instantiated in a void context. I expect to see something like `Capistrano::SCM::Git.new.setup!`. The reason is that I think of an initializer as something which, with respect to the surrounding state, returns something, not something that mutates surrounding state. So, I guess it looks to me like non-OOey magic, in a way. With that said, there are other cases of instantiating a class mutating world state, so this is really one man's opinion. (This and I'm not looking forward to telling other developers to not mutate world state on class instantiation and then have them point to Capistrano as an example of it doing it. :wink:)

So, overall, I'm very much for the underlying refactor, somewhat ambivalent on the configuration syntax change, and somewhat against the magic class instantiation.

I'd suggest just adding a `#setup!` (naming things is hard, alternative ideas are welcome. `#use!` or `#register!` might be better.) method to the class instantiation step.

I haven't gone through the code yet, I'll do so now.

lib/capistrano/configuration/scm_resolver.rb
```diff
@@ -0,0 +1,149 @@
...
+          register_legacy_scm_hooks
+          load_legacy_scm_by_name
+        end
+      end
+
+      private
+
+      def using_default_scm?
+        return @using_default_scm if defined? @using_default_scm
+        @using_default_scm = (fetch(:scm) == DEFAULT_GIT)
```
will-in-wi: Would this method be better as:
```ruby
def using_default_scm?
  @using_default_scm ||= (fetch(:scm) == DEFAULT_GIT)
end
```

Is the fetch and comparison slow enough to be worth caching? Or is the instance variable used anywhere else? If the latter, requiring that the method be called before you can do other things feels like a code smell.

lib/capistrano/configuration/scm_resolver.rb
```diff
@@ -0,0 +1,149 @@
...
+      end
+
+      # This is somewhat of a hack, because there is no guarantee that a third-
+      # party SCM will necessarily implement the typical SCM tasks. But it works
+      # well enough for the built-in SCMs.
+      def scm_plugin_loaded?
+        Rake::Task.tasks.any? { |t| t.name =~ /^[^:]+:create_release/ }
```
will-in-wi: Would it make more sense to require that scm plugins register themselves to avoid this? This feels like it could lead to some difficult to debug situations for a developer creating a new scm plugin.

lib/capistrano/configuration/scm_resolver.rb
```diff
@@ -0,0 +1,149 @@
...
+        fetch(:scm)
+      end
+
+      def load_built_in_scm
+        require "capistrano/scm/#{scm_name}"
+        scm_class = Object.const_get(built_in_scm_plugin_class_name)
+        # We use :load_immediately because we are initializing the SCM plugin
+        # late in the load process and therefore can't use the standard
+        # load:defaults technique.
+        scm_class.new(:load_immediately => true)
```
will-in-wi: Nitpick and point of curiosity: what's the Capistrano project's stance on Ruby 1.9 hash syntax? Does Cap support 1.8?

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: Or possibly throw a warning if `fetch(:scm)` is set and there are no such tasks.

lib/capistrano/configuration/scm_resolver.rb
```diff
@@ -0,0 +1,149 @@
...
+          after "deploy:new_release_path", "#{scm_name}:create_release"
+        end
+
+        if Rake::Task.task_defined?("deploy:check")
+          before "deploy:check", "#{scm_name}:check"
+        end
+
+        if Rake::Task.task_defined?("deploy:set_current_revision")
+          before "deploy:set_current_revision",
+                 "#{scm_name}:set_current_revision"
```
will-in-wi: For all of these, does it make sense to verify that scm_name:task_name actually exists?

lib/capistrano/configuration/scm_resolver.rb
```diff
@@ -0,0 +1,149 @@
...
+        elsif built_in_scm_name?
+          warn_set_scm_is_deprecated
+        elsif third_party_scm_name?
+          warn_third_party_scm_must_be_upgraded
+        end
+      end
+
+      def warn_set_scm_is_deprecated
+        $stderr.puts(<<-MESSAGE)
+WARNING: `set :scm, #{scm_name.inspect}` is deprecated.
```
will-in-wi: Would this be better as a NOTICE rather than a WARNING? I think most users would read this and think something broke...

lib/capistrano/scm/git.rb
```diff
@@ -0,0 +1,68 @@
...
+
+class Capistrano::SCM::Git < Capistrano::Plugin
+  def set_defaults
+    set_if_empty :git_shallow_clone, false
+    set_if_empty :git_environmental_variables, lambda {
+      {
+        :git_askpass => "/bin/echo",
+        :git_ssh => "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
+      }
+    }
```
will-in-wi: Personally, I find multiline blocks to be easier to read with do...end syntax rather than curly brace syntax. Not sure what the project standard is.

lib/capistrano/scm/git.rb
```diff
@@ -0,0 +1,68 @@
...
+    if (depth = fetch(:git_shallow_clone))
+      git :clone, '--mirror', '--depth', depth, '--no-single-branch', repo_url, repo_path.to_s
+    else
+      git :clone, '--mirror', repo_url, repo_path.to_s
+    end
+  end
+
+  def update_mirror
+    # Note: Requires git version 1.9 or greater
+    if (depth = fetch(:git_shallow_clone))
```
will-in-wi: Whenever I see this convention, I assume the author made a == mistake, and then realize that it is actually assignment. Would it make sense to pull this out?

lib/capistrano/scm/git.rb
```diff
@@ -0,0 +1,68 @@
...
+      git :fetch, '--depth', depth, 'origin', fetch(:branch)
+    else
+      git :remote, :update, '--prune'
+    end
+  end
+
+  def archive_to_release_path
+    if (tree = fetch(:repo_tree))
```
will-in-wi: Same comment per = convention

lib/capistrano/scm/tasks/git.rake
```diff
@@ -1,16 +1,7 @@
...
-  end
-
-  set :git_environmental_variables, ->() {
-    {
-      git_askpass: "/bin/echo",
-      git_ssh:     "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
-    }
-  }
+# This trick lets us access the Git plugin within `on` blocks.
+git = self
```
will-in-wi: This might be better named as git_plugin... git can mean a lot of things in here.

lib/capistrano/scm/tasks/hg.rake
```diff
@@ -1,23 +1,24 @@
-namespace :hg do
-  def strategy
-    @strategy ||= Capistrano::Hg.new(self, fetch(:hg_strategy, Capistrano::Hg::DefaultStrategy))
-  end
+# TODO: this is nearly identical to git.rake. DRY up?
 
+# This trick lets us access the Hg plugin within `on` blocks.
+hg = self
```
will-in-wi: hg_plugin?

lib/capistrano/scm/tasks/svn.rake
```diff
@@ -1,23 +1,24 @@
-namespace :svn do
-  def strategy
-    @strategy ||= Capistrano::Svn.new(self, fetch(:svn_strategy, Capistrano::Svn::DefaultStrategy))
-  end
+# TODO: this is nearly identical to git.rake. DRY up?
 
+# This trick lets us access the Svn plugin within `on` blocks.
+svn = self
```
will-in-wi: svn_plugin?

will-in-wi:
I'm done going through the code. I'm sure that a lot of my comments exhibit an ignorance of this project's conventions, so take them with a grain of salt.

Thank you both so much for your work on this! I'm excited to see this move forward.

will-in-wi:
Looking further into the code, I see how the plugin instantiation works. I guess I'd expect something like `Capistrano.register Capistrano::SCM::Git.new`, where something in the Capistrano DSL allows for the registration of plugins instead of registration being a function of the plugin itself. Make sense?

I don't know how old the plugin framework is, or whether it is feasible to change this before 4.0, even if desired.

Just my 2 cents.

mattbrictson:
Thanks for the thorough review! I'll respond in depth when I have more time.

Regarding the issue of the initializer having side-effects of mutating global state: I agree this is a code smell and it makes me uncomfortable. It also has the unfortunate consequence that if someone subclasses a plugin but forgets to call `super` in the initializer, then those effects are unintentionally skipped.

I was basing the behavior off of Rake, where e.g. initializing a task mutates the global state. In Rake there is a `Rake::TestTask` that does this, and other libraries follow suit, like RuboCop's `RuboCop::RakeTask`. These all make changes to the Rake environment as a side-effect of `new`. Since Capistrano is built on Rake, it seemed like a reasonable pattern to follow.

The more explicit approach, as you suggested would be:

```ruby
require "capistrano/scm/git"
Capistrano::SCM::Git.new.install!
```

And perhaps we could add a DSL helper so you could write instead:

```ruby
install "capistrano/scm/git"
```

mattbrictson:
> I don't know how old the plugin framework is, or whether it is feasible to change this before 4.0, even if desired.

The plugin system is brand new and has not yet been released, so feel free to propose sweeping changes. :wink:

will-in-wi:
My *personal* opinion is that this is a poor pattern and should not be continued. Having a plugin only implement API methods and having an external class be responsible for registering plugins is much more consistent with OO practice, I think. Sandi Metz once pointed out that inheritance is not a tool for code reuse, and that is kind of what it feels like here. It looks like the plugin superclass only exists to share the code for registration. That's probably its own concern which should be part of Capistrano core.

I'd rather see a pattern where you register a plugin by passing it a class constant which the registration class initializes. Part of the standard API would be to pass `self` to the plugin's constructor, which gives the plugin access to the rest of Capistrano. This allows plugins to be self-contained.

Does this make sense?

mattbrictson:
I see where you are coming from, but the reality as I see it is that Rake has already set the stage by making everything global.

When you write tests for a Rake task, for example, you don't mock or stub or pass in any "context" to the task, you just initialize it, which pollutes the global state. Then at the end of your test you call `Rake::Task.clear` to wipe the global state clean.

I worry that any OO formalities that we add are really just extra work for not much benefit, because the underlying system upon which Capistrano is based has already broken all those rules.

mattbrictson:
To summarize my thinking that led to the current plugin implementation: I would like to embrace Rake's conventions rather than fight them (even though much of Rake is counterintuitive to me).

will-in-wi:
Makes sense. Thanks for engaging in the discussion! At least a clear, thought-through decision has been made rather than just breaking the rules by inertia. :smile:

Does it make sense to at least have a `PluginClass.new.register!` method or something? Class instantiation should not, I think, mutate outside of itself.

Other than that, this sounds reasonable, and definitely an improvement.

I wish I could ask Jim Weirich why he decided to design Rake the way he did…

leehambley:
Moved to milestone 3.5 since the 3.5 release includes new plugins that already use the plugin architecture - so it'd be nice to back that out, or have a decision. 

FWIW, not related here, but since 3.5 is a nice round number I have an emotional attachment to wanting to get the SCM stuff pulled into a plugin, finally. (I'll pick up that thread though in that ticket)

will-in-wi:
One other thought I had as I've pondered this for a bit: Why not just have the `require` instantiate the plugin? The extra step of instantiating a class in a void context feels like a hat-tip to OO design, which is unnecessary unless we go the rest of the way. The instantiation cannot be done without the require, and the require is useless without the instantiation. Therefore, we have two statements which always come in pairs (even in the tests, it seems); why not just do the one?

Does this make sense? Heh, this is the type of discussion which happens better over beers. :smile: Thanks for the work on all of this!

mattbrictson:
> One other thought I had as I've pondered this for a bit: Why not just have the `require` instantiate the plugin?

This is an elegant solution for the 90% case, but what if you want to subclass it or call `new` with non-defaults? If you can't `require` without having side-effects then it severely limits extensibility.

mattbrictson:
This is a good discussion, btw! I feel like there is a better solution out there that we just haven't figured out yet. :smiley:

will-in-wi:
Makes sense. I guess I end up thinking that having an API like:

```ruby
require 'plugin_name'
Capistrano::Plugins.register PluginName.new
```

Is the most reasonable API that avoids mutating global state within a constructor. It also has the benefit of being Rubyish and reading nicely: "Capistrano plugins registers a new instance of my plugin." `Capistrano::Plugins` could also be an instance of something instantiated by a previous include...

mattbrictson:
I think we all agree that the ideal user experience is a simple `require` and nothing else. This is already what people are used to in Capistrano (e.g. `require "capistrano/rails"`).

Can we make that work and still have these benefits?

* Ability to subclass a plugin
* Load a plugin without its hooks (i.e. parameterizing the plugin registration process)

(You might argue that the OO pattern of subclassing to override behavior is not Rake-like and we should promote a different technique. But let's set that aside for now.)

Ruby has an `inherited` callback that let's us know when a class is subclassed. Capistrano could use this as a registry of sorts, keeping track when subclasses of `Plugin` are defined via `require`. This gets us registry behavior without needing an explicit `Capistrano::Plugins.register` step, and allows us to move registration code out of the plugin initializer. Con: it is a bit "magical".

Then there is the subclassing problem. If I want to build on top of Capistrano's Git plugin by subclassing it, both the original Git plugin and the subclass need to be loaded via `require`. But we only want the subclass to be registered. How does the implicit registration system handle this? It might be possible by walking the inheritance tree and registering only "leaf" classes, but that is even more magic.

On the final point: let's say I want to load a plugin but I want to opt out of its hooks. Maybe that is where explicit registration is acceptable, since it is an edge case:

```ruby
require "myplugin"
Capistrano::Plugins.register Myplugin, :hooks => false
# or in the current style, it would be
Myplugin.new(:hooks => false)
```

Anyway, now that I've written it all out, does it sounds like too much magic? Should I put together a proof of concept so we have actual code to critique?

mattbrictson:
BTW, the `inherited` technique is how Rails does it: https://github.com/rails/rails/blob/v4.2.5.1/railties/lib/rails/railtie.rb#L129-L133

will-in-wi:
That would be a really clever solution. With that said, I'm guessing that it would prove brittle, because we have to anticipate and code around every possible edge case. For example, what if someone wanted to subclass Git but use both the parent and subclass for some reason?

The way I look at it, you have two audiences: users who are just trying to get their apps to deploy, and developers who are trying to write maintainable code and are used to the language paradigms.

Users will copy/paste whatever the readme of a plugin tells them to. A single-line registration isn't any different in my mind to class instantiation.

Developers will want a low barrier of entry, which is accomplished by keeping the interfaces as standard as possible for the language. If all they have to do is implement a class which implements specific, documented methods, all of their prior knowledge is used without having to learn too much new stuff.

If we use some kind of magic solution, users will be unaffected, and developers will be basically learning a new object paradigm.

Capistrano is really interesting in that it uses Ruby as its DSL. It isn't writing a new language parser or using some kind of declarative configuration file, it just hands you Ruby. Based on this, how far should we go in exposing Ruby vs. hiding it? If we expose Ruby, we should probably try and follow Ruby OO paradigms. If we hide Ruby, we shouldn't try too hard to take advantage of its power, aka inheritance.

With all of this said, these are my personal opinions. I'm happy to provide feedback as someone who writes Ruby (and a couple of other languages) for a living and uses Capistrano every day. But you guys have put way more time into this and it is much more your project than mine.

Does this make sense? Please let me know what level of participation you want from me. You originally asked for my opinion on the SCM to new plugin system PR, and this somehow became a conversation around the new plugin API. If this is unwelcome, please let me know! I appreciate all of your work on this project - it makes my life much easier every day.

mattbrictson:
Very well said! I appreciate your feedback and time you are putting into this.

I'll work next on a PR based on our discussion here to refine the Plugin API, and then we can later circle back to the SCM stuff.

mattbrictson:
@leehambley Can we push the SCM redesign to a 3.6 milestone?

The 3.5.0 release is already a large one, and includes what will be a significant change to the default output (pretty → airbrussh). I'm not sure I want to add a wordy SCM deprecation notice/warning to that mix. Better to package it in its own release as 3.6?

Also there is a lot of code review left to do in this PR, and I don't want to delay 3.5 further.

leehambley:
I completely missed that you had already done this work, I thought it was still needing to be started. Moved to 3.6 milestone

mattbrictson:
@leehambley @will-in-wi Having let this PR settle for a few months and looking at it with fresh eyes, I am having doubts that this is the right approach. In hindsight, the complexity of `SCMResolver` should have been a red flag. I think this is an indication that the approach is flawed.

Ultimately this PR is affects a very large chunk of code and introduces deprecations that will annoy users for not much benefit. Yes, it will make a "none" SCM (or more accurately, a lack of SCM) possible, and may make SCM customizations easier in the long run, but what percentage of the Capistrano user base needs that? (I, for one, am happy with the default Git implementation.)

I'm considering closing this PR, or at least putting it on hold indefinitely.

leehambley:
> I'm considering closing this PR, or at least putting it on hold indefinitely.

It's still one of the more frequent complaints that we get, that SCMs are not pluggable. Not to mention the near impossibility of catering to the cacophony of potential different ways to combine Git and it's settings (modules, tree, clone, archive, fetch, shallow, etc)

Let's not lose this, I'd prefer to keep it open, even it just an anchor, and record of just what thought and work has gone into this, it might help to guide us to an implementation we can live with.

jjoos:
Ah that's sad, I was anticipating this change getting through and already started using that branch. (which works nicely). Anyway to help out getting this in master?

mattbrictson:
@leehambley @jjoos Thanks for the encouragement. I will try to get this cleaned up and ready for merge today. 👊

mattbrictson:
OK, this ready for final review! I made the following changes:

* Change the deprecation messages from `WARNING` to `[Deprecation Notice]` so that they look less severe (per @will-in-wi's recommendation).
* Rebased and squashed down to one big commit.
* Added a CHANGELOG entry.
* Added a new UPGRADING.md doc that explains how existing projects can be upgrading to silence the deprecations.
* Added a link to the SCM plugin documentation (see corresponding PR here: https://github.com/capistrano/documentation/pull/188)

All tests pass, including Cucumber tests.

@leehambley Does this look good to merge?

@jjoos You said you had been using this branch and it was working for you. Can you test again with the latest changes?

Thanks!

will-in-wi:
👍 Thanks for working on this!

lib/capistrano/templates/Capfile
```diff
@@ -4,6 +4,17 @@ require "capistrano/setup"
...
+#
+# require "capistrano/scm/hg"
+# Capistrano::SCM::Hg.new
+# or
+# require "capistrano/scm/svn"
+# Capistrano::SCM::Svn.new
+# or
+require "capistrano/scm/git"
+Capistrano::SCM::Git.new
+
```
will-in-wi: Should these be `install_plugin Capistrano::SCM::Git` now?

mattbrictson: Oh wow, I missed that. Good catch!

mattbrictson: Fixed.

leehambley:
I've a block of time set out 2016-05-24 to look through all the issues @mattbrictson and I'll make sure to give this one last lookover then, but basically yes, thanks for taking the time.

mattbrictson:
Note that this PR removes `:git_strategy`. My assumption was that the "strategy" mechanism of the old SCM implementation was not really used. However a recent `cap doctor` report from an user clued me in that this is actually being used in the wild.

For example [capistrano-git-submodule-strategy](https://rubygems.org/gems/capistrano-git-submodule-strategy) is under active development and has 56,000 downloads.

I'm not feeling good about merging this PR if it breaks such "strategy plugins" with no migration path.

leehambley:
I agree, but this strategy plugin should *replace* our Git module, that was the intent behind making the SCM pluggable... since then you wouldn't need to completely gut the core with a new "strategy" and could load whichever maintained Git/Hn/copy/rsync/none plugin you wanted....

Let's leave it open a bit longer, maybe @ekho wants to weigh in (plugin maintainer)

ekho:
Could you open issue in my repo?

leehambley:
hey @ekho welcome to the 🎉 !

We're not as far as opening issues yet, we've labelled this for discussion... what is your take on the whole thing?

ekho:
I think we can merge my plugin into Git plugin and manage behaviour with variable. 
Like 
set :git_use_submodules, true

mattbrictson:
As previously discussed in #1409, we'd like to avoid adding more conditionals like `:git_use_submodules`. What we're considering in this PR is using a new plugin system for these kinds of specialized SCM cases.

The idea is that instead of writing a "strategy" module, you would probably subclass the Git plugin, like this:

```ruby
class Capistrano::SCM::Git::WithSubmodules < Capistrano::SCM::Git
  # Override the default implementation as needed
end
```

Reference: https://github.com/mattbrictson/capistrano/blob/2aaaf54e042d621556d746b93401abb6c05102f2/lib/capistrano/scm/git.rb

Or if there is very little in common with the built-in Git plugin, you could start from a clean slate:

```ruby
class Capistrano::SCM::Git::WithSubmodules < Capistrano::Plugin
  # Implement whatever tasks you see fit
end
```

Reference: https://github.com/mattbrictson/capistrano.github.io/blob/8914b058a9f360c5f908e5dbdda07f2dab22db98/documentation/advanced-features/custom-scm/index.markdown

@ekho Is this something you'd be interested in trying out?

ekho:
@mattbrictson I created an issue in my repo. When this pr will be merged I will make changes in my plugin.

mattbrictson:
I've updated the CHANGELOG in this PR to make it clear that `:git_strategy`, etc. have been removed.

> The `:git_strategy`, `:hg_strategy`, and `:svn_strategy` settings have been
  removed with no replacement. If you have been using these to customize
  Capistrano's SCM behavior, you will need to rewrite your customization using
  the [new plugin system](https://github.com/capistrano/documentation/pull/188).

@leehambley @ekho Should we perhaps issue an interim release of Capistrano with deprecation warnings for `:git_strategy`, etc. using variables validation? That will at least alert people that changes are coming that will require they rewrite their SCM customizations.

mattbrictson:
Note to self: make sure to retain the fix from #1742 when resolving merge conflicts.

will-in-wi:
What's the status of this? I'm in the process of building a new deployment for my business and found a use case which would make this very helpful. Is there anything I can do to help get it out?

mattbrictson:
I think this is ready to go, pending a rebase. I'll take care of that and then ask you for a final review. Thanks for the bump!

mattbrictson:
- [x] Rebased
- [x] Feature specs pass
- [x] #1742 reapplied

@leehambley @will-in-wi Can you both do a final review and give a 👍 so I can merge? Thanks!

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: Unimportant, but still relevant, I think.

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: Is `scm_name` guaranteed to exist?

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: Still worth discussing, methinks.

lib/capistrano/scm/git.rb
will-in-wi: Still possibly valid.

lib/capistrano/scm/git.rb
will-in-wi: and here

lib/capistrano/configuration/scm_resolver.rb
mattbrictson: This is simply replicating the existing behavior, which you can see in the master branch here:
https://github.com/capistrano/capistrano/blob/3d4b17c2fb10291d0a31adbd8b3cd3a086f3c9bf/lib/capistrano/tasks/deploy.rake#L228

In master, it assumed that if you use `set :scm, "foo"`, then `foo:create_release`, `foo:check`, and `foo:set_current_revision` *must* exist. I am maintaining that behavior here when the resolver uses "legacy mode".

lib/capistrano/configuration/scm_resolver.rb
mattbrictson: What I am trying to accomplish here is testing whether `set :scm` was ever called. I do that by initializing `:scm` to a dummy value (`DEFAULT_GIT`) and then testing it like this:

```ruby
# if true, then :scm wasn't changed by the user
fetch(:scm) == DEFAULT_GIT
```

However, for the rest of the SCM code to work, and to not break backwards compatibility with other code that is using the `:scm` value, I have to change `:scm` to equal `:git` later on in the code flow, here: https://github.com/capistrano/capistrano/pull/1572/files#diff-f1f4ccb7067eef4a3d33b71bf7ecc2f8R29

That is why I record the value of the test in `@using_default_scm`, because `fetch(:scm)` will change later. I agree this is a hack, but I couldn't think of a better approach.

Any ideas?

lib/capistrano/configuration/scm_resolver.rb
mattbrictson: Yeah, this is tricky. I guess registration is unavoidable. How should SCM plugins register themselves? Do you think all SCM plugins should inherit from a `Plugin::SCM` parent class, perhaps? Or add metadata to the plugin that describes what it does, like `provides :scm`?

lib/capistrano/scm/git.rb
mattbrictson: I guess this is a matter of style. The extra parens are supposed to be a clue to the reader that the single `=` is intentional. In other words:

```ruby
# This is a mistake: it should be ==
if tree = fetch(:repo_tree)

# This is intentional; parens say "I meant to do that"
if (tree = fetch(:repo_tree))
```

However if this is confusing to other maintainers, then that is unequivocally a bad thing and I should break it into multiple lines for clarity.

mattbrictson:
Note to self: make sure https://github.com/capistrano/capistrano/commit/cb2ee06d2036b5906c5dfe8027238b1dc05a8875 is not lost in the merge.

mattbrictson:
@will-in-wi I added some replies to your line comments but they aren't showing up directly in the PR, so maybe they got lost. Did you have a chance to read them?

Also, I'm wondering if we can merge this PR now and deal with the remaining issues as a separate PR? Maintaining this branch is turning out the be a lot of work due to the file moves; lots of conflicts whenever master changes.

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: I think I was thinking that

```ruby
return @using_default_scm if defined? @using_default_scm
@using_default_scm = (fetch(:scm) == DEFAULT_GIT)
```

Was identical to

```ruby
@using_default_scm ||= (fetch(:scm) == DEFAULT_GIT)
```

But it technically isn't if `@using_default_scm == nil`.

It does turn this into a little bit of a state machine (or rather a quantum machine because state is set at time of observation. 😄 ), but this isn't so bad.

Disregard this comment.

lib/capistrano/configuration/scm_resolver.rb
will-in-wi: Can we take advantage of the `Capistrano::Git.new` stanza to have it register itself? IIRC, elsewhere in this change, it tracks things that were called that way…

lib/capistrano/scm/git.rb
will-in-wi: This was addressed elsewhere. Don't worry about it.

lib/capistrano/scm/git.rb
will-in-wi: This sounds reasonable. I don't personally (usually) use this style, but I'm not bothered by it as long as it is consistent.

will-in-wi:
Hey, sorry for taking so long on this. I started responding a while ago and realized that I didn't understand this deeply enough at the time. I still don't, but I'd rather move it along.

I made a couple of comments. My main functional question would be around plugin registration instead of assuming the existence of a rake task. However, I'm totally fine throwing that into a different ticket.

LGTM!!! 👍

mattbrictson:
Rebased, feature tests are passing. Finally merging this in!! 🎉🎉🎉

mattbrictson:
I've moved the outstanding issues here: #1792, #1793.

will-in-wi:
Thank you so much for your work on this!

ekho:
@mattbrictson I created new plugin for submodules support - [capistrano-git-with-submodules](https://github.com/ekho/capistrano-git-with-submodules)

PR #1546, opened 2015-12-29
Added option to set specific revision when using Subversion as SCM

marcovtwout:

PR #1573, opened 2016-01-25
Added option to set specific revision when using Subversion as SCM

marcovtwout:
Squashed https://github.com/capistrano/capistrano/pull/1546


mattbrictson:
I don't use Subversion so I can't really test this, but it looks good. Thanks!

mattbrictson:
@marcovtwout BTW, next time I'd prefer doing this all in one PR. You can rebase your branch locally and then using `git push --force` to update the remote, and the PR will automatically update to reflect the new commits. That way you don't have to close and open a new PR just to do a rebase, and we don't lose track of the conversation.

This article has a good writeup on how to do it: https://www.sammyk.me/how-to-contribute-to-an-open-source-project-on-github

PR #1578, opened 2016-02-05
Fix host filter regression

### LINKED ISSUE #1577: Hosts filter with regex has regressed in master
From https://github.com/capistrano/documentation/pull/169

Capistrano 3.4.0 behavior is:

    cap --hosts=^local production deploy

Capistrano master behavior is:

    cap --hosts=/^local/ production deploy

The alternatives to not appear to work.

cshaffer:
Fixes #1577 

When I broke out the different filter strategies in #1479 I thought I could collapse the regex behavior into what I thought was a lowest common denominator based on the test suite alone, but I was wrong.

This PR adds a test for the case I definitely broke, and puts back all the regex stuff the way it was in case there are other cases that regressed.


mattbrictson:
Excellent! :sparkles:

PR #1580, opened 2016-02-11
Remove JRuby from the README

mattbrictson:
Although JRuby might work, we don't officially support it (i.e. not tested by Travis), so best take it out of the official docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Capistrano is designed to deploy using a single, non-privileged SSH user, using 
 
 ### Requirements
 
-* Ruby version 2.0 or higher on your local machine (MRI, JRuby, or Rubinius)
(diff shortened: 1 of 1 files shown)


will-in-wi:
Dumb question: is there any reason that jruby isn't tested by Travis?

leehambley:
It doesn't work. Some things we take for granted (hash access being atomic, iirc) don't work in jruby and neither matt nor myself use it.

There's also some underlying problem with Net::SSH via SSHKit which causes hard to track down bugs, so we stopped supporting jRuby. (all as far as I can remember, some of this might not be true anymore)

will-in-wi:
Thanks for the explanation! Sounds good.

PR #1585, opened 2016-02-22
Prune old tracking branches from the git repository on each host

javanthropus:
This change runs a git prune operation prior to running the repository update operation in order to remove old tracking branches.  Updates for projects that have a large branch turnover rate grind to a fairly sudden halt without regular pruning as the server side of the repository becomes bogged down trying to resolve branch refs that no longer exist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ https://github.com/capistrano/capistrano/compare/v3.4.0...HEAD
 * Drop support for Ruby 1.9.3 (Capistrano may still work with 1.9.3, but it is
   no longer officially supported)
 * Added new runtime option `--print-config-variables` that inspect all defined config variables in order to assist development of new capistrano tasks (@gerardo-navarro)
+* Prune dead tracking branches from git repositories while updating
+  * Git version 1.6.3 or greater is now required
 
 * Minor changes
   * Fix filtering behaviour when using literal hostnames in on() block (@townsen)
--- a/lib/capistrano/git.rb
+++ b/lib/capistrano/git.rb
@@ -34,7 +34,7 @@ class Capistrano::Git < Capistrano::SCM
       if (depth = fetch(:git_shallow_clone))
         git :fetch, '--depth', depth, 'origin', fetch(:branch)
       else
-        git :remote, :update
+        git :remote, :update, '--prune'
       end
     end
 
--- a/spec/lib/capistrano/git_spec.rb
+++ b/spec/lib/capistrano/git_spec.rb
@@ -61,7 +61,7 @@ module Capistrano
     describe "#update" do
       it "should run git update" do
         context.expects(:fetch).with(:git_shallow_clone).returns(nil)
-        context.expects(:execute).with(:git, :remote, :update)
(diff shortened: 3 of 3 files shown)


javanthropus:
Somehow the CI checks caused a segfault in Rubinius.  My changes are extremely minimal, so I'm not sure how they could have caused that.

mattbrictson:
Neat! I didn't know about this. Thanks for the nice PR. A couple things:

The git-prune manpage says:

>  Note: In most cases, users should run git gc, which calls git prune.

Do you agree that `git gc` or perhaps `git gc --auto` would be better than `git prune`? I don't have any experience with this, so I'll rely on your best judgment.

Second, is there any downside to this change that you can think of? My guess is that most Capistrano users are probably getting by just fine without pruning, so we need to be careful about making a change that has no upside for them but could cause problems. For example, would pruning on every deploy potentially make things *slower* for some people?

mattbrictson:
BTW, Travis is now :white_check_mark:. The segfault must have been a fluke.

javanthropus:
The command in this patch is different than `git prune`, actually, which is intended to remove objects in your repository that no longer have any references that can reach them, such as when you rewrite a commit and the prior commit object is left without anything referencing it anymore (aside from the reflog).  The `git remote prune` command is only used to remove tracking branches from a remote in the repository that no longer exist in the named remote repository.

From our experience, running `git remote prune` is pretty quick, but it obviously must reach out to the remote repository in order to determine which branches no longer exist.  If the remote repository is generally slow to respond, this will also be slow.

Since version 1.6.3 of git, there is an option for `git remote update` that tells git to perform a prune operation during the update.  We could use that in order to avoid adding additional latency of a separate prune operation assuming we're willing to cut off users who may still use such old git versions that do not support this option.

Unfortunately, I haven't actually tried using this option when we've run into the "too many dead branches" problem, so I can't say with certainty that using the `--prune` option would avoid the hangs in such cases.  I can test that out tomorrow though and report back if desired.  Otherwise, we can use the patch as implemented and avoid kicking out users of older git versions.  Let me know.

javanthropus:
I just realized that I introduced this by saying that this patch calls `git prune`, which is misleading.  Sorry for the confusion there.

mattbrictson:
Thanks for the detailed explanation. It looks like Git 1.6.3 was released 6 years ago; to me it doesn't seem unreasonable to require it. I would be interested in pursuing the `git remote update --prune` approach.

@leehambley Can I get a second opinion on this one?

leehambley:
Let's just do it, I don't know of any distro that is using Git 1.6 by default, and we can't wait forever :)

javanthropus:
Using the `--prune` option for `git remote update` seems to work well, so I've rewritten my patch to use that instead of a separate call to `git remote prune origin`.

mattbrictson:
Can you add something to the CHANGELOG to the effect of, "Capistrano now requires Git 1.6.3 or newer"?

javanthropus:
Added a note about the minimum git version to the CHANGELOG.

mattbrictson:
Thanks for the contribution! :raised_hands:

PR #1586, opened 2016-02-23
Allow more than one value to be sent to `append`

mattbrictson:
I was about to write documentation for the new `append` DSL method, when I realized that in practice you would very likely want to append more than one value at a time. For example:

```ruby
append :linked_dirs, ".bundle", "tmp"
```

/cc @kirs


kirs:
:+1: thanks!

PR #1587, opened 2016-02-24
Implementation of remove.

will-in-wi:
I just saw https://github.com/capistrano/capistrano/pull/1447 and realized how useful this is, especially when paired with the ability to remove. This PR adds that ability.

My use case is that I've built an internal Gem for my company which sets a pile of defaults. Sometimes, I don't need all of the default linked_dirs or such, at which point I have some messy code to remove them from the array.


leehambley:
@will-in-wi welcome to the fold committing to Cap itself :)

will-in-wi:
Thanks!

PR #1590, opened 2016-02-26
Clarify that Capistrano::Plugin is not fully baked

mattbrictson:
Given the ongoing discussion in #1572, I think the documentation for `Capistrano::Plugin` should be updated to discourage use until its API has stabilized.

/cc @leehambley @will-in-wi


will-in-wi:
:+1:

PR #1592, opened 2016-02-27
Move plugin installation logic out of initializer

mattbrictson:
Plugins are now installed via a `install_plugin` DSL method, rather than as a side-effect of calling `new`. You may pass either the plugin class or the instance of the class to `install_plugin`; both accomplish the same thing.

Example:

```ruby
require "capistrano/superfancy"
install_plugin Capistrano::Superfancy
```

The logic for executing the plugin is now encapsulated in the following class: `Capistrano::Configuration::PluginInstaller`.

This PR is based on the discussion in #1572.

/cc @will-in-wi @leehambley

### DIFF
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -1,5 +1,6 @@
 require_relative "configuration/filter"
 require_relative "configuration/question"
+require_relative "configuration/plugin_installer"
 require_relative "configuration/server"
 require_relative "configuration/servers"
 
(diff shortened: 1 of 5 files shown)


will-in-wi:
I like this a lot. My only comment is the name of the `hooks` param: would `load_hooks` be better? Trying to find a term that captures the idea of whether or not the hooks are autoloaded on installation.

mattbrictson:
Rebased to fix conflicts. Changed `hooks` to `load_hooks`.

mattbrictson:
OK, I'm dipping my toe back in to some Capistrano PRs. @will-in-wi @leehambley does this one look good to merge? I've rebased to fix the conflicts, and all Cucumber features pass.

leehambley:
Yes, absolutely

will-in-wi:
:+1: 

Thanks for your work on this!

PR #1593, opened 2016-02-28
Upgrade airbrussh dependency to 1.0.0 final

mattbrictson:
Airbrussh 1.0.0 is now released, so we no longer need to point to the beta version.


leehambley:
yikes, conflicts :(

mattbrictson:
Conflicts fixed.

Issue #1591, opened 2016-02-27
Rubocop?


will-in-wi:
Is there any interest in implementing a Rubocop ruleset for Capistrano? It might improve the consistency of PRs and such. I'm happy to take a first crack at it, but I wanted to check first.

Thanks!

mattbrictson:
I am a huge fan of RuboCop for new projects, but it gets a little tricky for existing ones. I suspect there will be many violations that we will have to ignore, or someone will have to do the tedious work of correcting them. If you want to take a crack at it, then by all means!

For Capistrano, we already have RuboCop included, but we're only running the lint checks at the moment (see .travis.yml and Rakefile).

In terms of the ruleset, I suppose we should look at the existing code to see what are the most common conventions, and go with that.

My personal preference is for hash rockets (controversial, I know) and double-quoted strings, but otherwise the RuboCop defaults. Here's my typical RuboCop config:

```yaml
AllCops:
  DisplayCopNames: true
  DisplayStyleGuide: true
  TargetRubyVersion: 1.9
  Exclude:
    - "*.gemspec"

Metrics/AbcSize:
  Exclude:
    - "test/**/*"

Metrics/MethodLength:
  Exclude:
    - "test/**/*"

Metrics/ClassLength:
  Exclude:
    - "test/**/*"

Style/ClassAndModuleChildren:
  Enabled: false

Style/DoubleNegation:
  Enabled: false

Style/HashSyntax:
  EnforcedStyle: hash_rockets

Style/SpaceAroundEqualsInParameterDefault:
  EnforcedStyle: no_space

Style/StringLiterals:
  EnforcedStyle: double_quotes

Style/TrivialAccessors:
  AllowPredicates: true
```

But, like I said, it probably makes sense to tweak the rules based on the patterns being used by the existing codebase, and not based on personal preference, unless those preferences are very strongly held (mine are not).

mattbrictson:
#1594 has been merged, but I would say another PR is still needed for:

1. Change the Rakefile so the default task runs the entire RuboCop suite, not just `--lint`.
1. Update `.travis.yml` to match.
1. Update the **Coding guidelines** section of `DEVELOPMENT.md`.
1. Update `CHANGELOG.md` if you want (not really necessary, since this doesn't effect end-users).

will-in-wi:
I agree, I'll get going on additional PRs soon.

will-in-wi:
[...]

PR #1594, opened 2016-02-28
Rubocop

will-in-wi:
Per https://github.com/capistrano/capistrano/issues/1591, here is a first crack at Rubocopifying the project. I've got a config, and everything passes. I tried to simply disable the cops that might be under discussion, and fix stuff that seemed mostly reasonable (to my personal definition of reasonable).

I started with the config @mattbrictson provided.

Discuss!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,3 +15,91 @@ Style/StringLiterals:
   EnforcedStyle: double_quotes
 Style/TrivialAccessors:
   AllowPredicates: true
+
+# Need discussion
+Style/AndOr:
+  Enabled: false
+Style/Documentation:
+  Enabled: false
+Style/RedundantSelf:
+  Enabled: false
+Style/RedundantReturn:
+  Enabled: false
+Style/SpaceAfterComma:
+  Enabled: false
+Style/ParallelAssignment:
+  Enabled: false
+Style/RegexpLiteral:
+  Enabled: false
+Style/SignalException:
+  Enabled: false
+Style/AccessorMethodName:
+  Enabled: false
+Performance/RedundantMerge:
+  Enabled: false
+Style/PercentLiteralDelimiters:
+  Enabled: false
+Style/BlockDelimiters:
+  Enabled: false
+Style/ConditionalAssignment:
+  Enabled: false
+Style/AlignParameters:
+  Enabled: false
+Style/IfUnlessModifier:
+  Enabled: false
+Style/Next:
+  Enabled: false
+Style/Lambda:
+  Enabled: false
+Style/EmptyLiteral:
+  Enabled: false
+Style/RaiseArgs:
+  Enabled: false
+Style/SingleLineBlockParams:
+  Enabled: false
+Style/Not:
+  Enabled: false
+Style/WhileUntilModifier:
+  Enabled: false
+Style/RescueModifier:
+  Enabled: false
+Style/SpecialGlobalVars:
+  Enabled: false
+Style/ModuleFunction:
+  Enabled: false
+Style/DotPosition:
+  Enabled: false
+Style/BlockEndNewline:
+  Enabled: false
+Style/AlignArray:
+  Enabled: false
+Style/MultilineBlockLayout:
+  Enabled: false
+Style/TrailingCommaInLiteral:
+  Enabled: false
+Style/Proc:
+  Enabled: false
+Style/CommandLiteral:
+  Enabled: false
+Style/AlignHash:
+  Enabled: false
+
+# Needs refactors
+Metrics/PerceivedComplexity:
+  Enabled: false
+Metrics/CyclomaticComplexity:
+  Enabled: false
+Metrics/MethodLength:
+  Enabled: false
+Style/PredicateName:
+  Enabled: false
+Metrics/LineLength:
+  Enabled: false
+Metrics/AbcSize:
+  Enabled: false
+Style/PerlBackrefs:
+  Enabled: false
+Metrics/ClassLength:
+  Enabled: false
+Metrics/ModuleLength:
+  Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,17 @@
+AllCops:
+  DisplayCopNames: true
+  DisplayStyleGuide: true
+  TargetRubyVersion: 1.9
+
+Style/ClassAndModuleChildren:
+  Enabled: false
+Style/DoubleNegation:
+  Enabled: false
+Style/HashSyntax:
+  EnforcedStyle: hash_rockets
+Style/SpaceAroundEqualsInParameterDefault:
+  EnforcedStyle: no_space
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
+Style/TrivialAccessors:
+  AllowPredicates: true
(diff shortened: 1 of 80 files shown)


will-in-wi:
If anyone wants to see the changes which would occur for a particular cop, use `bundle exec rubocop --only copname`.

My thoughts on the remaining cops:

## Style/AndOr
I prefer the `&&` and `||` styles to `and` and `or`, personally.

## Style/Documentation
I dislike this requirement, unless there is a bunch of autogenerated documentation.

## Style/RedundantSelf
Rubocop wants to nix any cases of `self.something` when `something` would work. I like this, but I can see why others might not.

## Style/RedundantReturn
This would favor Ruby's implicit return value, rather than using the explicit `return` keyword. I like this, although I personally follow the convention of having a newline before the implicit return.

## Style/SpaceAfterComma
I like this, but others may not. This is when you have method arguments like `|a,e|` or `(arg1,arg2)`, should there be a space after the comma.

## Style/ParallelAssignment
Rubocop dislikes cases like `@arg1, @arg2 = arg1, arg2`. I personally find these cases to be slightly confusing, but I also like their terseness. I could go either way on it.

## Style/RegexpLiteral
Some rules regarding how regexes are used. I prefer them.

## Style/SignalException
Prefer `raise` over `fail`. Ironically, this is a reversal of their previous stance. I just go with `raise` everywhere.

## Performance/RedundantMerge
Easiest explained by the output message:
```lib/capistrano/dsl/env.rb:29:11: C: Performance/RedundantMerge: Use names.last[:exclude] = :no_release instead of names.last.merge!(:exclude => :no_release). (https://github.com/JuanitoFatas/fast-ruby#hashmerge-vs-hash-code)```
Seems like a straightforward performance increase.

## Style/PercentLiteralDelimiters
These are rules for how various percent-literal usages should be delimited. For example, %r should always be used with (), not {}. I personally like this, but it doesn't matter much.

## Style/BlockDelimiters
This asks you to use `do...while` for multi-line blocks instead of `{...}`. I personally like this rule.

## Style/ConditionalAssignment
```lib/capistrano/tasks/console.rake:8:5: C: Style/ConditionalAssignment: Use the return of the conditional for variable assignment and comparison.``` I like this rule, very rubyish.

## Style/AlignParameters
This requires a little more cleanup, to keep legible, but I prefer the final product with this rule.

## Style/IfUnlessModifier
Forces the usage of the modified form of `if` and `unless` for short statements. I like this one.

## Style/Next
This encourages a guard clause in loops instead of wrapping the bulk of the loop body in a conditional. I like this, normally. Sometimes I prefer the alternative, but usually just go with the rule for consistency's sake.

## Style/Lambda
Suggests that we use the pointy stick (`->`) syntax over the `lambda` keyword syntax. I like this.

## Style/EmptyLiteral
Use hash literal {} instead of Hash.new. I like this. I see why others may not.

## Style/RaiseArgs
This encourages the `raise ErrorClass, 'error message'` syntax over the `raise ErrorClass.new('error message')` syntax.

## Style/SingleLineBlockParams
```lib/capistrano/configuration.rb:149:31: C: Style/SingleLineBlockParams: Name reduce block params |a, e|.```
I'm not sure what I think about this. It is nice that `reduce` instances always use the accumulator and element terminology. But sometimes it is nice to think in terms of the name of the thing in question. So is the Functional concept more important than the specific implementation?

## Style/Not
Asks that we always use `!` instead of `not`. I personally prefer this.

## Style/WhileUntilModifier
I've never run across this one before. In a single line while loop, you can use `while` as a modifier, much like `if`. Kind of cool.

## Style/RescueModifier
Rubocop recommends not using `rescue` as a modifier. I didn't know you could. I personally think that this is just a form of swallowing errors, which I would discourage, so I like this rule.

## Style/SpecialGlobalVars
This encourages us to use more descriptive names from the English module in the standard library. E.g. `$CHILD_STATUS` over `$?`. I like this, since I always forget the Perl-like special chars.

## Style/ModuleFunction
Use module_function instead of extend self. Not sure what this entails.

## Style/DotPosition
In a multi-line method chain, place the dot at the beginning of each line, rather than at the end. I prefer this.

## Style/BlockEndNewline
Encourages that multi-line block terminators always reside on their own lines. I prefer this, particularly when Style/BlockDelimiters is enabled.

## Style/MultilineBlockLayout
For a multi-line array, Rubocop doesn't like when the first element is on the same line as the block definition. I prefer Rubocop's approach.

## Style/TrailingCommaInLiteral
Avoid comma after the last item of an array. I generally just go with Rubocop on this one, although I think I prefer having a comma on the last line, mostly because it doesn't mess with diffs as much when you add additional lines. I'd prefer that it be enforced one way or the other.

## Style/Proc
Prefer `proc` over `Proc.new`. I agree.

## Style/CommandLiteral
Prefer backticks over whatever is being used right now. I have no idea what the impact of this is.

## Style/AlignHash
Align the elements of a hash literal if they span more than one line. I like this.

will-in-wi:
If it would be preferred, we can evaluate the PR as it stands now, and then open new tickets for enabling other cops.

We probably want to at least take a first pass and turn on the ones we all easily agree on.

will-in-wi:
Further comments. I grouped the disabled cops into the ones which need discussion and the ones which need refactors. At some point (probably in a different ticket) it should be discussed whether we should turn some of them on and either fix or disable the cops for the cases which break now.

Personally, I prefer the Ruby 1.9 hash syntax and single quotes. :smile: But I'm not going to worry about that one unless others agree with me.

will-in-wi:
Oh, and I tried to make each cop's fixes into its own commit, so it would be easy to rewrite history and revert a cop that we actually don't want.

mattbrictson:
Excellent work! I want to wait for @leehambley's input before merging or moving ahead with this.

In the meantime, I generally agree with all of the RuboCop defaults for the remaining cops that you listed, with a few additional comments:

* Style/Documentation: I would like to be able to enable this one day. Perhaps not in this PR.
* Style/ParallelAssignment: I agree with RuboCop that parallel assignment should be avoided.
* Style/PercentLiteralDelimiters: This has annoyed me enough times that I'd be happy to turn it off.
* Style/SingleLineBlockParams: I don't like this one.
* Style/ModuleFunction: There are subtle differences in behavior between the `module_function` and `extend self` styles. We should be careful about applying this rule, or perhaps we should disable it, because changes here could introduce tricky bugs.

mattbrictson:
I suspect the hash rockets and quotes will be the ones that warrant further discussion, as they have the biggest impact in terms of number of lines affected.

With that in mind, could you make the following changes to this PR to get it ready to merge?

1. Remove the hash-rockets and string literals commits (perhaps move them to a separate PR).
2. Change the Rakefile so the default task runs the entire RuboCop suite, not just `--lint`.
3. Update `.travis.yml` to match.
4. Update the **Coding guidelines** section of `DEVELOPMENT.md`.
5. Update `CHANGELOG.md` if you want (not really necessary, since this doesn't effect end-users). 

Then I think this PR will be good to go, assuming @leehambley doesn't raise any objections. After that we can tackle the more controversial cops in separate PRs.

leehambley:
My only objection is that there's *anything* to discuss about syntax and style. One of the better things about Go is a canonical way to format the code, which the *compiler* enforces. I'm glad to see us pick a fairly sane split. Thank you both.

PR #1595, opened 2016-03-01
Rubocop housekeeping

will-in-wi:
Followup to https://github.com/capistrano/capistrano/pull/1594.

* Updates TravisCI to use full Rubocop suite
* Updates docs to reflect change


mattbrictson:
:ship:

PR #1596, opened 2016-03-01
Use default Rubocop recommended syntax for string literals.

will-in-wi:
For discussion.

This changes the default syntax to single quotes unless there is string interpolation or apostrophes.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,8 +11,6 @@ Style/DoubleNegation:
   Enabled: false
 Style/SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
-Style/StringLiterals:
-  EnforcedStyle: double_quotes
 Style/TrivialAccessors:
   AllowPredicates: true
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,10 @@
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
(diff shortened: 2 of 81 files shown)


mattbrictson:
I prefer double quotes, and I think the trend among open source is going in this direction. My unscientific survey of the popular Ruby projects on GitHub shows a preference for double-quotes. And FWIW, thoughtbot is on board as well: their HoundCI code style service defaults to double quotes.

will-in-wi:
I think it will be easier for me to keep track of whack-a-mole, if you guys don't mind. I just need some sort of official approval on each PR, and I'll rebase and clear it for merge.

will-in-wi:
I personally prefer single quotes. My argument for it is because it allows me at a glance to see whether a string is doing anything different like string interpolation.

With that said, this really is personal preference. It wouldn't bother me at all if Capistrano stuck with double-quotes.

will-in-wi:
Okay, last one. I've rebased, but it sounds like this might be a close as wontfix.

Let's make a call one way or the other and close or merge. @leehambley ?

will-in-wi:
Well, let's call that a decision. Closing.

Thanks for the discussion!

leehambley:
Ohh, I was in favour of merge :) but no strong opinion.

will-in-wi:
Heh. I don't care much. I go back and forth on this particular one. Let's stick with double-quotes.

PR #1598, opened 2016-03-01
Enable Rubocop default for AndOr syntax

will-in-wi:
For discussion.


mattbrictson:
It is easy to make mistakes with `and` and `or` because of their precedence rules, so I prefer this. :+1:

will-in-wi:
Looks like the TravisCI linter changes broke this one, so I've rebased. Awaiting results...

will-in-wi:
Tests passing.

PR #1599, opened 2016-03-01
Enable redundant self cop

will-in-wi:
For discussion.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,8 +21,6 @@ Style/AndOr:
   Enabled: false
 Style/Documentation:
   Enabled: false
-Style/RedundantSelf:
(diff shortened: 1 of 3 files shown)


mattbrictson:
It can sometimes be useful to explicitly refer to `self` just to call attention, but as a general rule, I think this is fine.

will-in-wi:
I agree on the whole.

I've personally come to like this rule because it forces you to be conscious of the few times when `self.something` and `something` are actually different. On codebases to which I've applied this rule, I now can see a `self.*` and immediately know that it is unique.

leehambley:
Yep, one of my pet peeves of Ruby programmers who `self.` everything because they don't know the difference!

PR #1601, opened 2016-03-01
Enable SpaceAfterComma cop

will-in-wi:
For discussion.


mattbrictson:
Seems like an easy win for readability. :+1:

PR #1603, opened 2016-03-01
Enable RegexpLiteral cop

will-in-wi:
For discussion.


mattbrictson:
I'm mildly in favor of this. At least it is a good idea to standardize on `/.../` syntax.

PR #1604, opened 2016-03-01
Enable SignalException cop

will-in-wi:
For discussion.

https://github.com/capistrano/capistrano/pull/1594


mattbrictson:
Sure. :+1:

will-in-wi:
Rubinius doesn't like this for some reason...

will-in-wi:
Rebase and kick solves the Rubinius issue.

PR #1606, opened 2016-03-01
Enable BlockDelimiters cop.

will-in-wi:
For discussion.


mattbrictson:
I don't have a reference handy, but I know some people like to use `{ }` vs `do end` based on the *intention* of the block (procedure vs predicate, maybe?) rather than the *length* of the block. I use Rubocop's suggestion in my own code, but I don't have a strong preference on this one.

PR #1607, opened 2016-03-01
Enable ConditionalAssignment cop

will-in-wi:
For discussion.

### DIFF
--- a/lib/capistrano/tasks/console.rake
+++ b/lib/capistrano/tasks/console.rake
@@ -5,11 +5,11 @@ task :console do
   loop do
     print "#{stage}> "
 
-    if (input = $stdin.gets)
-      command = input.chomp
-    else
-      command = "exit"
-    end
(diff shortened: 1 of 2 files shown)


lib/capistrano/tasks/console.rake
```diff
@@ -5,11 +5,11 @@ task :console do
...
-    if (input = $stdin.gets)
-      command = input.chomp
-    else
-      command = "exit"
-    end
+    command = if (input = $stdin.gets)
+                input.chomp
+              else
+                "exit"
+              end
```
mattbrictson: In this particular case, maybe this is better?

```ruby
command = (input = $stdin.gets) ? input.chomp : "exit"
```

will-in-wi:
I agree per the ternary statement. Change made.

mattbrictson:
This cop is fine by me. :+1:

PR #1609, opened 2016-03-01
Enable the IfUnlessModifier cop

will-in-wi:
For discussion.


mattbrictson:
One of the nice features of Ruby. :+1:

PR #1610, opened 2016-03-01
Enable Style/Next cop

will-in-wi:
For discussion.


mattbrictson:
It took a little getting used to, but since RuboCop nudged me toward this style I've grown to like it. :+1:

PR #1613, opened 2016-03-01
Enable RaiseArgs cop

will-in-wi:
For discussion.


mattbrictson:
I don't really understand RuboCop's reasoning for this one, but I have no problem with it.

PR #1614, opened 2016-03-01
Enable Style/Not cop

will-in-wi:
For discussion.


mattbrictson:
I didn't even know Ruby had a `not` operator. :blush: Looks good. :+1:

PR #1616, opened 2016-03-01
Enable RescueModifier cop

will-in-wi:
For discussion.

### DIFF
--- a/features/support/remote_command_helpers.rb
+++ b/features/support/remote_command_helpers.rb
@@ -16,7 +16,9 @@ module RemoteCommandHelpers
   end
 
   def safely_remove_file(_path)
-    run_vagrant_command("rm #{test_file}") rescue VagrantHelpers::VagrantSSHCommandError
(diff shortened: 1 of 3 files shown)


features/support/remote_command_helpers.rb
```diff
@@ -16,7 +16,9 @@ def exists?(type, path)
   end
 
   def safely_remove_file(_path)
-    run_vagrant_command("rm #{test_file}") rescue VagrantHelpers::VagrantSSHCommandError
+    run_vagrant_command("rm #{test_file}")
+  rescue
+    VagrantHelpers::VagrantSSHCommandError
```
mattbrictson: This looks like a bug. My guess is that the intention was

```ruby
rescue VagrantHelpers::VagrantSSHCommandError
  nil
end
```

will-in-wi:
Good catch, fixed.

mattbrictson:
I think this is a case in point that inline `rescue` is a bad idea. :+1:

PR #1617, opened 2016-03-01
Enable SpecialGlobalVars cop

will-in-wi:
For discussion.

### DIFF
--- a/capistrano.gemspec
+++ b/capistrano.gemspec
@@ -3,8 +3,6 @@ lib = File.expand_path("../lib", __FILE__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require "capistrano/version"
 
-require "English"
-
 Gem::Specification.new do |gem|
   gem.name          = "capistrano"
   gem.version       = Capistrano::VERSION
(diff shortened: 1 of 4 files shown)


mattbrictson:
I like this, but we have to be very careful to `require "English"` or it doesn't work. Perhaps RSpec or another dependency is doing it already, which is why tests pass. But without the `require` it may break for actual `cap` use.

capistrano.gemspec
```diff
@@ -12,7 +12,7 @@ Gem::Specification.new do |gem|
   gem.summary       = "Capistrano - Welcome to easy deployment with Ruby over SSH"
   gem.homepage      = "http://capistranorb.com/"
 
-  gem.files         = `git ls-files`.split($/)
+  gem.files         = `git ls-files`.split($INPUT_RECORD_SEPARATOR)
```
mattbrictson: This might be safer, and doesn't require `English`: `git ls-files -z`.split("\x0")

will-in-wi:
I added the English module to affected files.

will-in-wi:
I also modified the gemspec per comment.

PR #1618, opened 2016-03-01
Enable DotPosition cop

will-in-wi:
For discussion.


mattbrictson:
Very much agree with this one. :+1:

PR #1622, opened 2016-03-01
Enable Style/Proc cop

will-in-wi:
For discussion.


mattbrictson:
I don't see any downside to this. Shorter == better. :+1:

PR #1624, opened 2016-03-01
Enable AlignHash cop

will-in-wi:
For discussion.


mattbrictson:
No strong opinion either way on this one.

PR #1625, opened 2016-03-02
Enforce %Q syntax for string literals (disallow bare %)

mattbrictson:
Here's one more RuboCop suggestion.

By default, RuboCop prefers using the bare `%` for declaring string literals that contain both single and double quotes, like this:

```ruby
%{Isn't this "great".}
```

`%Q` does the same thing, but it is more explicit:

```ruby
%Q{Isn't this "great".}
```

I prefer the latter syntax. Many times I have made the mistake where I mean to declare an array using `%w`, but I forget the `w` and end up with a `%` string instead. It is a subtle typo and can be easy to miss.

```ruby
%(one two three)  # Looks like an array but it's not
%w(one two three) # What I meant to write
```

Prohibiting the use of bare `%` helps prevent this mistake.

/cc @will-in-wi

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,8 @@ AllCops:
   DisplayStyleGuide: true
   TargetRubyVersion: 2.0
 
+Style/BarePercentLiterals:
+  EnforcedStyle: percent_q
 Style/ClassAndModuleChildren:
   Enabled: false
 Style/DoubleNegation:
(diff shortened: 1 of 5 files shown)


mattbrictson:
Here's the Stack Overflow question that brought this to mind: http://stackoverflow.com/a/35732693/4625365

will-in-wi:
Oooo, good catch in the SO question! I approve whole-heartedly. :+1:

leehambley:
Wow, I spent 20min with him on IRC, sent him to SO in the end, absolutely
approved.

mattbrictson:
Haha, OK that settles it, then! :smile:

PR #1638, opened 2016-03-25
Do not abbreviate git revision

schneiderderek:
There doesn't seem to be any significant benefit in forcing the revision SHA to be a shorter length. 

@z5h brings up some good points here: https://github.com/capistrano/capistrano/pull/1469
and here: https://github.com/capistrano/capistrano/pull/1423

If someone needs the abbreviated version of the revision, then they can always shorten it themselves. It's a lot harder to get the full SHA out of an abbreviated version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ Other changes:
   * Allow dot in :application name (@marcovtwout)
   * Fixed git-ssh permission error (@spight)
   * Added option to set specific revision when using Subversion as SCM (@marcovtwout)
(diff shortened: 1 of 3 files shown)


leehambley:
Please check the contribution guidelines @schneiderderek

schneiderderek:
@leehambley I read through the contribution guidelines more carefully and added an entry to the changelog.

leehambley:
Thanks @schneiderderek :)

jesperronn:
Update, sorry, I found the PR, and now understand the rationale :)

PR #1653, opened 2016-04-27
Explain that recent change to `after` hook resolution could break deploys

### LINKED ISSUE #1652: document 3.5.0 namespace-aware hooks change as potentially breaking
one of my cap files

```ruby
namespace :app do
  task :restart do
    # restarts the app server
  end

  after 'deploy:publishing', :restart
end
```

After upgrading to 3.5.0 my deploy silently ignored the `after` directive.
Had to change it to `after 'deploy:publishing', 'app:restart'`

mattbrictson:
Fixes #1652 

/cc @leehambley @dreyks


leehambley:
@mattbrictson for a pure doc change are you fine with not re-releasing? I just ask because [`CHANGELOG.md` is part of the Gem bundle](https://github.com/capistrano/capistrano/blob/master/capistrano.gemspec#L15) a..

mattbrictson:
I'm fine with not re-releasing.

PR #1654, opened 2016-04-27
Enable syntax highlight

thiagotalma:


mattbrictson:
Thanks for the PR, but this is not a change I'm comfortable merging. IMHO highlighting configuration is better done in the editor itself. I know Sublime Text for example has the ability to associate file names and extensions with certain syntaxes, so that Gemfile, Rakefile, Procfile, Capfile, etc. can be interpreted as Ruby. I'm sure other editors have similar abilities.

leehambley:
~~Thanks for the PR, we don't have a policy against these kinds of commits, but I can't accept this one as it introduces modelines for some editors, and probably not others.~~

Sniped by@mattbrictson! I agree completely, thanks for picking that up Matt :)

Issue #1670, opened 2016-05-10
Reported bad build because of Rubocop


leehambley:
Rubocop has apparently gone mad, the change to the README marks the build as broken.

* https://travis-ci.org/capistrano/capistrano/builds/129168416

The outputs there seem new.

will-in-wi:
I wonder whether there is a Rubocop version increase...

will-in-wi:
Yeah, master breaks too. Fixing...

mattbrictson:
Fixed by #1671

PR #1671, opened 2016-05-10
Fix new rubocop linter issues

will-in-wi:
Fix issue in https://github.com/capistrano/capistrano/issues/1670

### DIFF
--- a/lib/capistrano/configuration/filter.rb
+++ b/lib/capistrano/configuration/filter.rb
@@ -10,11 +10,10 @@ module Capistrano
       def initialize(type, values=nil)
         raise "Invalid filter type #{type}" unless [:host, :role].include? type
         av = Array(values)
-        @strategy = case
-                    when av.empty? then EmptyFilter.new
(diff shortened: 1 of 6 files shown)


lib/capistrano/configuration/filter.rb
```diff
@@ -10,11 +10,10 @@ class Filter
...
         av = Array(values)
-        @strategy = case
-                    when av.empty? then EmptyFilter.new
-                    when av.include?(:all), av.include?("all") then NullFilter.new
-                    when type == :host then HostFilter.new(values)
-                    when type == :role then RoleFilter.new(values)
+        @strategy = if av.empty? then EmptyFilter.new
+                    elsif av.include?(:all) || av.include?("all") then NullFilter.new
+                    elsif type == :host then HostFilter.new(values)
+                    elsif type == :role then RoleFilter.new(values)
```
will-in-wi: > Style/EmptyCaseCondition: Do not use empty case condition, instead use an if expression

mattbrictson:
Excellent work. Thanks!

PR #1672, opened 2016-05-11
Make name of current directory configurable

websi:
Add configuration variable to define name of directory/symlink "current"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
   * Make path to git wrapper script configurable (@thickpaddy)
   * Change git wrapper path to work better with multiple users (@thickpaddy)
+  * Make name of current directory configurable via configuration variable `:current_directory` (@websi)
 
 ## `3.5.0`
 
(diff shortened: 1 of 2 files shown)


mattbrictson:
What's your use case for changing the `current` name? Just curious.

websi:
Some of our customers use a different directory structure for their Apache vhosts and they will not change this structure.
So we have to change the name of the "current" directory / symlink.
In this case, the name of the "current" directory / symlink is "htdocs".

will-in-wi:
To throw in my 2¢, when I've run into this scenario, I manually make `htdocs` a symlink to `current`.

In my experience, such cases usually also have other constraints which make Capistrano a difficult sell, so deploying Capistrano to a different directory entirely, and symlinking to that directory, has worked better for me.

mattbrictson:
I see. This seems like a fairly benign change that could make it easier/possible for some people to use Capistrano, so I'm willing to merge it.

But first, can you make the following edits?

1. Fix merge conflicts.
2. In your commit message, mention the "htdocs" scenario that motivated this change, just so that future contributors running `git blame`, etc. will understand what this feature is for.
3. In the CHANGELOG entry, mention the directory name can be overridden by setting `:current_directory`. As it stands the bullet point is a little vague.

Thanks!

websi:
Updated the pull request

mattbrictson:
Thanks! 🙌

Issue #1656, opened 2016-05-05
Ruby 1.9.3 issue with new-style arguments in method definition used in Capistrano::Configuration


tadman:
This uses the newer Ruby 2.x syntax for defining arguments:

    def install_plugin(plugin, load_hooks:true)

#### Steps to reproduce

1. Install Capistrano 3.5.0 with Ruby 1.9.3-p551
2. Run `cap production deploy` or equivalent.
3. Error.

#### Expected behaviour

Should work if 1.9.3 support is still a priority. This locks at >= 2.0.0

#### Actual behaviour

Errors out:

    capistrano-3.5.0/lib/capistrano/configuration.rb:157: syntax error, unexpected keyword_end, expecting $end

Capistrano 3.4.1 does not have this issue, and Ruby 2.0.0 or better is fine with it.

mattbrictson:
Thanks for the detailed report!

In fact we dropped support for Ruby 1.9 in the Capistrano 3.5.0 release. It is mentioned in the [CHANGELOG](https://github.com/capistrano/capistrano/blob/master/CHANGELOG.md#350).

Once we removed 1.9 from our testing matrix I suppose it was inevitable that we would end up breaking 1.9, but to be honest I didn't expect it to happen this quickly. Sorry about that!

I'll close this issue at some point because 1.9 is no longer supported, but in the meantime, I will accept the following PRs if you (or other 1.9 users out there) would like to contribute to the project:

1. [Update our gemspec](http://guides.rubygems.org/specification-reference/#required_ruby_version=) to enforce the >= 2.0.0 requirement. Newer versions of Bundler [will honor this requirement](http://bundler.io/blog/2015/12/12/version-1-11-released.html#new-features), so there will be less of a chance that a 1.9 user inadvertently ends up with a broken Capistrano.
2. Modify our CHANGELOG to indicate that Capistrano 3.5.0 *definitely* no longer works with 1.9.3 (as opposed to *may* no longer work).

mattbrictson:
Closed via #1678

PR #1678, opened 2016-05-15
Explicit requirement of Ruby >= 2.0, reflection of that in the changelog.

irvingwashington:
This is a PR as suggested in #1656.


mattbrictson:
Excellent, thank you for contributing to the project!

PR #1682, opened 2016-05-17
Raise a better error when an ‘after’ hook isn’t found

jdelStrother:
Prior to this, if an after-hook didn’t exist, we’d end up calling `nil.invoke`, which would end up in the DSL `invoke` method and raise “wrong number of arguments (0 for 1+)”

WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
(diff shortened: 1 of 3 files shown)


mattbrictson:
Great improvement to the friendliness of the tool, and a spec as well. Thanks! ❤️ 

Could you add a CHANGELOG entry so I can merge this in? You can credit yourself in the CHANGELOG message if you want.

jdelStrother:
Sure, updated.

PR #1683, opened 2016-05-17
WIP: Making the default :application validator accept blocks.

### LINKED ISSUE #1681: Setting :application to a Proc causes cap to crash

#### Steps to reproduce

#### Expected behaviour

#### Actual behaviour

```
cap aborted!
```

irvingwashington:
I'm in doubt regarding tests for this. Ideally I'd test the validator on it's own, but it also makes sense to test the `set :application, -> { "some_name" }` case integrated. 
Please let me know what you think, and I'll update this PR.

(work in progress) Fixes #1681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Reverse Chronological Order:
 ## master
 
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
-
+  * Make the default :application validator handle blocks correctly (@irvingwashington)
   * Raise a better error when an ‘after’ hook isn’t found (@jdelStrother)
   * Restrict the uploaded git wrapper script permissions to 700 (@irvingwashington)
   * Make path to git wrapper script configurable (@thickpaddy)
--- a/lib/capistrano/defaults.rb
+++ b/lib/capistrano/defaults.rb
@@ -1,4 +1,6 @@
-validate :application do |_key, value|
+validate :application do |_key, value_or_block|
+  value = value_or_block.respond_to?(:call) ? value_or_block.call : value_or_block
+
   changed_value = value.gsub(/[^A-Z0-9\.\-]/i, "_")
   if value != changed_value
     warn %Q(The :application value "#{value}" is invalid!)


mattbrictson:
Good start, but the more I think about this, the more it seems that validation has a core design flaw that will be hard to fix.

The whole point of passing a Proc to `set :application, ...` is to delay execution of that block. But validation is always executed immediately. And if the validation logic needs to call the Proc in order to check the resulting value, then we've completely defeated the purpose of delayed execution.

Perhaps we could solve this by delaying validation until `fetch` is called, instead of validating it on `set`.

Ultimately I am not convinced validation (as currently implemented) is a great feature; I think it has caused more bugs that it has solved. Perhaps it is better suited to be done during `cap doctor` so it doesn't interfere directly with `set` and/or `fetch`.

But I will let others weigh in.

/cc @leehambley @Kriechi

leehambley:
Hey @mattbrictson sorry to quote one sentence only out of what you wrote, I agree with the rest, regarding:

> Ultimately I am not convinced validation (as currently implemented) is a great feature; 

I don't think it introduced *bugs* per-se, but it did reduce what is allowed in that field, I like the outcome (e.g forcing the application name to be a valid dir name), here's my thoughts on the validation stuff in general:

 * Agreed, taking a lambda was only ever intended as a band-aid for allowing deferred execution, are there other use-cases we're not seeing here?
 * Agreed that the validation as-is would break that by forcing (at the naïve end, a `to_s` to be called very early)
 * Not sure if it's wise to do validation as part of the "resolution" of the variable (although we could move the validation to a lambda that evaluates the variable including validation the first time it is used)
 * Probably it makes sense to include validation in cap-doctor, but also then it's way, way, way too late if someone is running into issues, and we *could* have caught simple problems like bad variable names at runtime... but instead it's in another command (and we have no idea how often that's used, if at all, given that it's quite new)

mattbrictson:
> Agreed, taking a lambda was only ever intended as a band-aid for allowing deferred execution, are there other use-cases we're not seeing here?

I suppose someone could use `ask :application`, which is essentially a form of delayed execution. It is unlikely that you would want to prompt for that particular variable, but validation could be used for other prompted variables down the road.

> Probably it makes sense to include validation in cap-doctor, but also then it's way, way, way too late if someone is running into issues

Agreed.

leehambley:
Let's see if anyone comes along with a use-case other than delayed evaluation when procs are passed to `set()`, I can't see any cases, yet - but maybe there are.

bogdanRada:
Is there any chance for this being resolved? I have been using this gem https://github.com/railsware/caphub for quite a long time to have all the deployment scripts for all applications in a single application but this needs to set the aplication as follows:

```
set :application, proc { fetch(:stage).split(':').reverse[1] }
```

this is because the stage is formed from the application name followed by two dots (":") and then the name of the stage. 

I see there are still discussions if this should be allowed. In case though you decide to not allow this anymore, then what would be a workaround? 

I really like the new version 3.5.0 of capistrano, but this issue is currently a roadblock for me to update my application to the new version.

mattbrictson:
@bogdanRada Thanks for the comment. I agree that the 3.5.0 behavior is definitely a bug, and I'm sorry this is a roadblock for you.

Your use case also makes it clear that immediately calling the Proc for validation purposes is not a good solution. We need to either disable validation for Procs entirely, or implement a more full-featured solution that delays validation until the first time `fetch` is called.

This PR seems to be stalled. If you are able to propose a PR of your own we would appreciate the help.

mattbrictson:
Based on the discussion above, I created an alternative PR that implements delayed validation for procs: #1700.

mattbrictson:
Closed in favor of #1711.

PR #1685, opened 2016-05-18
Use git_wrapper_path when setting GIT_SSH

thickpaddy:
Apologies, this went missing when resolving merge conflicts for #1517 

@mattbrictson could you test this for me? Sorry, not using Capistrano at the moment, and don't have anything set up to test locally. Thanks.

### DIFF
--- a/lib/capistrano/tasks/git.rake
+++ b/lib/capistrano/tasks/git.rake
@@ -13,7 +13,7 @@ namespace :git do
   set :git_environmental_variables, lambda {
     {
       git_askpass: "/bin/echo",
-      git_ssh: "#{fetch(:tmp_dir)}/#{fetch(:application)}/git-ssh.sh"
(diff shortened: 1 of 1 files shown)


thickpaddy:
BTW I've dug out the original commit to confirm that this is all I missed, and it certainly seems to be. Should be good after this PR is merged, but apologies again for messing up all the same :-(

mattbrictson:
Thanks for the quick turnaround! I must have tested it before the merge conflicts were resolved and failed to notice the regression.

I'm running the Cucumber suite now and will do a test deployment after that.

mattbrictson:
Looks good. Thank you! 🙌

thickpaddy:
Woot! Thanks for testing 👍

leehambley:
Thank you both for a smooth fix 🚀

Issue #1684, opened 2016-05-18
Feature: display server/role data in doctor output


mattbrictson:
Capistrano's `server` and `role` DSL can be confusing. In some cases the options from multiple declarations are merged into a single server, but not always. It is easy to make mistakes when specifying role names. And, crucially, there is no easy feedback mechanism to see what this DSL has actually effected, other than to run a `deploy` and check for problems.

Many issues have been raised regarding this trouble: #1579, #1400, #1434, #1431, #1427.

I suggest enhancing the `cap doctor` output to include a report of all servers and their attributes.

Something like this (please feel free to recommend something better!):

```
Servers (3)

    deployer@10.0.0.1:522  [:app, :web]
    deployer@10.0.0.2      [:db, :redis]
    deployer@10.0.0.3      [:sidekiq]     { :custom => "properties" }
```

The important points to include are:

* The number of servers
* The hostname
* Username and port if they were explicitly specified
* The roles
* Any custom properties that were explicitly specified
* Perhaps highlight/print a warning if a role name contains spaces

mattbrictson:
Implemented by #1687

PR #1687, opened 2016-05-19
[wip] doctor:servers task

irvingwashington:
This is a basic version of a doctor task that shows configured servers along with their roles & options.

There are some internal api changes (public access to `Configuration#servers` and a new public `Server#to_h` method).

This version doesn't display any warnings/highlighting yet. Original issue #1684 mentions a warning about roles with whitespaces. I didn't find any definite indication that this is problematic (it probably requires some more research from my side), I'll add it if needed.
I'll run through stack overflow capistrano problems, to look for inspiration for warnings in this task, but in the meantime this basic version can be reviewed. 

Preview
![Feature preview]
(http://i.imgur.com/hZ5AyNd.png)

### DIFF
--- a/lib/capistrano/doctor/servers_doctor.rb
+++ b/lib/capistrano/doctor/servers_doctor.rb
@@ -1,0 +1,105 @@
+require "capistrano/doctor/output_helpers"
+
+module Capistrano
+  module Doctor
+    class ServersDoctor
+      include Capistrano::Doctor::OutputHelpers
+
+      def initialize(env=Capistrano::Configuration.env)
+        @servers = env.servers.to_a
+      end
+
+      def call
+        title("Servers (#{servers.size})")
+        rwc = RoleWhitespaceChecker.new(servers)
+
+        table(servers) do |server, row|
+          sd = ServerDecorator.new(server)
+
+          row << sd.uri_form
+          row << sd.roles
+          row << sd.properties
+          row.yellow if rwc.any_has_whitespace?(server.roles)
+        end
+
+        if rwc.whitespace_roles.any?
+          warning "\nWhitespace detected in role(s) #{rwc.whitespace_roles_decorated}. " \
+            "This might be a result of a mistyped \"%w()\" array literal."
+        end
+        puts
+      end
+
+      private
+
+      attr_reader :servers
+
+      class RoleWhitespaceChecker
+        attr_reader :whitespace_roles, :servers
+
+        def initialize(servers)
+          @servers = servers
+          @whitespace_roles = find_whitespace_roles
+        end
+
+        def any_has_whitespace?(roles)
+          roles.any? { |role| include_whitespace?(role) }
+        end
+
+        def include_whitespace?(role)
+          role =~ /\s/
+        end
+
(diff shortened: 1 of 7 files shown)


mattbrictson:
Wow, this is excellent work! Looks good as is, once a CHANGELOG entry is added.

Regarding the whitespace warning: this came up at least once before in an issue or in Stack Overflow. Ah, here it is: http://stackoverflow.com/questions/35718268/rails-assets-arent-compiling-after-capistrano-deployment/35732693#35732693

So it would be nice to have, especially since Capistrano users not familiar with Ruby syntax might easily mistake `%w{}` and `%{}`.

Also, what do you think of sorting the servers by hostname? Might be nice for deployments with dozens of servers.

lib/capistrano/doctor/servers_doctor.rb
```diff
@@ -0,0 +1,70 @@
...
+        def initialize(server)
+          @server = server
+        end
+
+        def uri_form
+          [
+            server.user,
+            server.user && "@",
+            server.hostname,
+            server.port && ":",
```
mattbrictson: I really like this `value && separator` shorthand you used here in combination with `.compact.join`. Elegant!

irvingwashington: Thanks!

irvingwashington:
Thanks for the feedback! That SO thread makes the case obvious indeed.
I've added this whitespace check, and the changelog entry.

Please let me know if the message is ok (eg. `Whitespaces detected in role(s) :"app web". This might be a result of a mistyped "%w()" array literal.`).

Preview
![Feature preview]
(http://i.imgur.com/0TV1nlh.png)

lib/capistrano/doctor/servers_doctor.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      class RoleWhitespaceChecker
+        attr_reader :whitespace_roles, :servers
+
+        def initialize(servers)
+          @servers = servers
+          @whitespace_roles = find_whitespace_roles
+        end
+
+        def have_whitespaces?(roles)
```
mattbrictson: I think `any_has_whitespace?(roles)` would be a better name for this method.

lib/capistrano/doctor/servers_doctor.rb
```diff
@@ -0,0 +1,105 @@
...
+          sd = ServerDecorator.new(server)
+
+          row << sd.uri_form
+          row << sd.roles
+          row << sd.properties
+          row.yellow if rwc.have_whitespaces?(server.roles)
+        end
+
+        if rwc.whitespace_roles.any?
+          warning "\nWhitespaces detected in role(s) #{rwc.whitespace_roles_decorated}. " \
```
mattbrictson: Grammar: I would say `"\nWhitespace detected` (no "s").

mattbrictson:
Looks great! See my two minor code comments.

irvingwashington:
Fixed and pushed.

mattbrictson:
Nice! ✨

Issue #1431, opened 2015-05-15
Duplicate Output in log debug pretty


leevigraham:
I've seen this lots of mentions of this without a clear solution.

I'm experiencing duplicate output in the log when running a staged deployments.

Here's my log:

```
INFO [abd4b2ba] Running /usr/bin/env mkdir -p /tmp/take5/ as newismco@eepau1-02.nexcess.net
INFO [abd4b2ba] Finished in 0.510 seconds with exit status 0 (successful).
INFO Uploading /tmp/take5/git-ssh.sh 100.0%
INFO [94e1f157] Running /usr/bin/env chmod +x /tmp/take5/git-ssh.sh as newismco@eepau1-02.nexcess.net
INFO [94e1f157] Finished in 0.090 seconds with exit status 0 (successful).
INFO [8c55a5f6] Running /usr/bin/env mkdir -p /tmp/take5/ as newismco@eepau1-02.nexcess.net
INFO [8c55a5f6] Finished in 0.102 seconds with exit status 0 (successful).
INFO Uploading /tmp/take5/git-ssh.sh 100.0%
INFO [8ef265f0] Running /usr/bin/env chmod +x /tmp/take5/git-ssh.sh as newismco@eepau1-02.nexcess.net
INFO [8ef265f0] Finished in 0.075 seconds with exit status 0 (successful).
Please enter branch (master): 
INFO [bd0e8af3] Running /usr/bin/env git ls-remote --heads git@github.com:newism/take5.git as newismco@eepau1-02.nexcess.net
INFO [bd0e8af3] Finished in 3.681 seconds with exit status 0 (successful).
INFO [2e1529c6] Running /usr/bin/env git ls-remote --heads git@github.com:newism/take5.git as newismco@eepau1-02.nexcess.net
INFO [2e1529c6] Finished in 3.685 seconds with exit status 0 (successful).
INFO [4ed7024d] Running /usr/bin/env mkdir -p ~/take5.ncig.com.au/shared ~/take5.ncig.com.au/releases as newismco@eepau1-02.nexcess.net
INFO [4ed7024d] Finished in 0.077 seconds with exit status 0 (successful).
[...]

PR #1688, opened 2016-05-21
Raise if a stage has an invalid name like "deploy"

mattbrictson:
Implement a sanity check as discussed here: https://github.com/capistrano/capistrano/issues/1431#issuecomment-219875383

I added a Cucumber feature to verify it works.

I didn't add a CHANGELOG entry since this is for a rare edge case that won't impact most users.

### DIFF
--- a/features/stage_failure.feature
+++ b/features/stage_failure.feature
@@ -1,0 +1,9 @@
+Feature: Stage failure
+
+  Background:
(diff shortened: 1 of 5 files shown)


mattbrictson:
@leehambley Any objections to merging this in?

leehambley:
:)

PR #1696, opened 2016-06-01
add customizable variables doctor

ncreuschling:
Variable `:doctor_variables_whitelist` (array of symbols, defaults to emtpy array) can now be set to customize additional variables for variables doctor

### DIFF
--- a/lib/capistrano/doctor/variables_doctor.rb
+++ b/lib/capistrano/doctor/variables_doctor.rb
@@ -12,7 +12,8 @@ module Capistrano
 
       include Capistrano::Doctor::OutputHelpers
 
-      def initialize(env=Capistrano::Configuration.env)
+      def initialize(additional_variables_whitelist=%i(), env=Capistrano::Configuration.env)
+        @additional_variables_whitelist = additional_variables_whitelist
         @env = env
       end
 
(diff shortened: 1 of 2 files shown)


mattbrictson:
Thanks for the PR! The doctor feature is brand new, so we're still trying to understand how it is used and how well it is working. What led you to create the `:doctor_variables_whitelist` feature? Is there something about the standard doctor behavior that's not working for your use case?

ncreuschling:
Thank you for starting this conversation.

Currently I am sitting on a collection of Capistrano extensions for deploying TYPO3 (my employer plans to release this into the public). Without this feature most of our variables will appear in the unrecognized section of the VariablesDoctor. This might confuse users.

Much like you I see this as an experiment. We have some other ideas for doctors as well and we could easily distribute this feature our own doctors for example. Please let me know if you find this idea interesting (I would add some tests then) and how we should proceed with this idea.

BTW, I have no clue why the build with rbx fails with a segmentation failt. It works for rbx-2.5.8 on my MacBook.

mattbrictson:
Don't worry about the rbx segfault; I don't think that has anything to do with this PR.

Its great to hear that you're considering adding doctors of your own. If you think it would be useful to make part of Capistrano core, you can follow this recent example of a PR that added a new doctor: https://github.com/capistrano/capistrano/pull/1687

Regarding the original issue: I designed the `doctor:variables` task so that it would build a whitelist automatically as long as certain patterns are followed:

* Any variables that your extensions use should have default values
* Those default values should be set using `set_if_empty` inside the `load:defaults` task
* Alternatively you can package your extensions as a [Capistrano::Plugin](https://github.com/capistrano/capistrano/blob/master/lib/capistrano/plugin.rb), in which case the plugin API will guide you into following these patterns

The idea is that `doctor:variables` implicitly "trusts" any variables that are set during Capistrano initialization process, which includes `load:defaults`. So as long as your extension sets its default values during that setup phase, those variables will be automatically whitelisted and you won't need `:doctor_variables_whitelist`.

That's the intention of the design, anyway. Does it sound like it is a good fit for your use case?

ncreuschling:
> That's the intention of the design, anyway. Does it sound like it is a good fit for your use case?

Thank you for your explanation. I will chose a solution and close this PR for now.

Issue #1694, opened 2016-05-29
Doctor incorrectly reports ":git_strategy is not a recognized Capistrano setting"


mattbrictson:
The `:git_strategy` variable is known by Capistrano and used by the built-in Git SCM. So this warning from `doctor:variables` is wrong:

```
:git_strategy                Capistrano::Git::SubmoduleStrategy

:git_strategy is not a recognized Capistrano setting (config/deploy.rb:22)
```

mattbrictson:
If someone wants to try fixing this, it should just be a matter of adding `:git_strategy, :hg_strategy, :svn_strategy` to `VariablesDoctor::WHITELIST`.

ShaneSaww:
@mattbrictson I have a commit ready to go that fixes this issue, but when setting up a new laptop I messed my git config. I'm going to fix that then re-upload this commit and open a PR once everything looks good. Assuming that works for you.

mattbrictson:
Thanks! The commit looks good and I would appreciate the PR once you are ready. That said, please see if you can also include a test and a CHANGELOG entry in the PR. I can give you pointers on writing the test if you get stuck.

ShaneSaww:
@mattbrictson Here is the new commit(if you want to make sure everything is good). 
I started working on the test, but I'm not 100% on how to best test it, so any guidance would be helpful.

mattbrictson:
OK, can you please open a PR with what you have so far? We can move the discussion there.

Don't worry about making the commit perfect, we can always make tweaks and clean things up before the PR is ultimately merged. The important thing is to have a PR so there is a natural place for the discuss and review.

ShaneSaww:
Awesome! PR is opened.

PR #1699, opened 2016-06-08
Updating the variables in the VariblesDoctor WHITELIST.

ShaneSaww:
This PR fixes: #1694 by adding `:git_strategy, :hg_strategy, :svn_strategy` to the Whitelist for VariablesDoctor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * Make path to git wrapper script configurable (@thickpaddy)
   * Change git wrapper path to work better with multiple users (@thickpaddy)
   * Make name of current directory configurable via configuration variable `:current_directory` (@websi)
-  * Updating the variables in the VariblesDoctor WHITELIST. (@shanesaww)
+  * `doctor` no longer erroneously warns that `:git_strategy` and other SCM options are "unrecognized" (@shanesaww)
 
 ## `3.5.0`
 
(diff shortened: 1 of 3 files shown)


mattbrictson:
Thanks! I'll provide some feedback once I have time to review.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * Make path to git wrapper script configurable (@thickpaddy)
   * Change git wrapper path to work better with multiple users (@thickpaddy)
   * Make name of current directory configurable via configuration variable `:current_directory` (@websi)
+  * Updating the variables in the VariblesDoctor WHITELIST. (@shanesaww)
```
mattbrictson: You've explained what was changed, but not why. I think this would be better:

> `doctor` no longer erroneously warns that `:git_strategy` and other SCM options are "unrecognized"

mattbrictson:
For the test, my recommendation is as follows.

Add `set :git_strategy, "some value"` to the `before` block here:

https://github.com/capistrano/capistrano/blob/1035c973b005b76346f764764b11ddef119ff94c/spec/lib/capistrano/doctor/variables_doctor_spec.rb#L15-L22

And then write a test that the doctor *does not* print a warning for it. Much like this test does:

https://github.com/capistrano/capistrano/blob/1035c973b005b76346f764764b11ddef119ff94c/spec/lib/capistrano/doctor/variables_doctor_spec.rb#L53-L57

```ruby
expect { doc.call }.not_to output(...).to_stdout
```

Does that make sense?

ShaneSaww:
Thanks! This all makes sense, I will make the updates and repush!

ShaneSaww:
@mattbrictson I think this is ready to go!

mattbrictson:
Thank you very much! 🙇

Issue #1702, opened 2016-06-16
update_asset_mtimes doesn't find manifest


vpuzzella:
Version 2.15.8
#### Steps to reproduce
1. run cap deploy:assets:update_asset_mtimes
#### Expected behaviour

Should successfully find manifest file and capture content
#### Actual behaviour

Does not find manifest file. I believe the bash command is incorrect. The wildcard should be placed outside of the quoted file path, like so:

```
      manifest_content = capture("[ -e '#{shared_path.shellescape}/#{shared_assets_prefix}/#{asset_manifest_prefix}'* ] && cat #{shared_path.shellescape}/#{shared_assets_prefix}/#{asset_manifest_prefix}* || echo").strip
```

mattbrictson:
Thanks for the report! Generally speaking we no longer support Capistrano 2.x, so I suggest testing out your proposed fix with your own fork of the project.

We'll consider accepting a PR for 2.x if you can make a strong case that this is a critical bug.

vpuzzella:
Patch: https://github.com/capistrano/capistrano/pull/1703

mattbrictson:
> IMO, it's pretty critical because asset mtimes are not being touched. That causes deploy:assets:clean_expired to remove required assets.

So everyone using 2.x to deploy a Rails app is seeing all of their assets being deleted? If so I would think this would be widely reported. Is this a recent regression?

vpuzzella:
Hard to say. Might be platform related (centos in my case). Or, not many people use the deploy:assets:clean_expired task. In may case, I added `after "deploy:restart", "deploy:cleanup"` which triggers ```deploy:assets:clean_expired```.

vpuzzella:
Here is a simple test I ran on my centos box:
```
$ [ -e 'README*' ]; echo $?
1
$ [ -e 'README'* ]; echo $?
0
```

mattbrictson:
OK, thanks for taking the time to explain. Can you do some more digging as to why the code was written the way it was? For example, it seems that your change will undo the fix for https://github.com/capistrano/capistrano/issues/578 (see also commit https://github.com/vpuzzella/capistrano/commit/57633b211f4fa260c2a85e98c2558497c4ac4fb4).

Please understand that I have no way of testing v2, so I have to be very careful about accepting PRs that have even a slight chance of introducing bugs or regressions.

vpuzzella:
[...]

PR #1703, opened 2016-06-16
Fixed bash command which tests for existence of manifest file.

vpuzzella:
Fixes https://github.com/capistrano/capistrano/issues/1702

### DIFF
--- a/lib/capistrano/recipes/deploy/assets.rb
+++ b/lib/capistrano/recipes/deploy/assets.rb
@@ -88,7 +88,7 @@ namespace :deploy do
     DESC
     task :update_asset_mtimes, :roles => lambda { assets_role }, :except => { :no_release => true } do
       # Fetch assets/manifest contents.
-      manifest_content = capture("[ -e '#{shared_path.shellescape}/#{shared_assets_prefix}/#{asset_manifest_prefix}*' ] && cat #{shared_path.shellescape}/#{shared_assets_prefix}/#{asset_manifest_prefix}* || echo").strip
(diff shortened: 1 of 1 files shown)


will-in-wi:
Are we still maintaining v2?

mattbrictson:
I have no way of testing v2, so for me personally, I am not maintaining it. Unless the PR fixes a critical bug in a way that is obvious from looking at the diff and has no chance of unintended consequences. This PR doesn't qualify IMHO.

leehambley:
I think I the patch makes sense, Bash won't expand a * in single quotes,
however I'm in favour of closing lest we send a signal that such small
issues qualify as "critical" for V2 releases.

will-in-wi:
Can we say that officially we do not support v2? Then we close all v2 related tickets, and if someone really wants to maintain v2, they can fork it, or perhaps we could work with them. But right now, no maintainers use it, and I think 3 has been out long enough to fix most of the issues and encourage migration.

I'd be in favor of wholly unsupporting v2.

mattbrictson:
I agree that v2 should be officially unsupported. Closing all v2 tickets seems severe, but it is probably the right thing to do, considering those issues will never be fixed.

*Unofficially,* we could make an exception to this rule in an extreme scenario. For example (very contrived): say a new version of Ruby or Bundler comes out and completely breaks v2, and the fix is a simple one-liner. I would hate to leave all v2 legacy users out in the cold just on principle.

leehambley:
Agreed, let's formally EOL V2 with the next release of cap we can mention
it in the release notes?

will-in-wi:
Here's one way of saying this in the CHANGELOG: https://github.com/capistrano/capistrano/pull/1932

will-in-wi:
We've officially stopped support of Capistrano 2 and are not accepting patches for it any longer.

Thank you for your work on this! Sorry for the long wait and unsatisfying conclusion.

PR #1704, opened 2016-06-17
Fix specs according changes in sshkit 1.11

lebedev-yury:
SSHkit 1.11 adds a `:known_hosts` option to `ssh_options`,
which breaks one spec in the current master.


leehambley:
Thanks, good catch!

PR #1706, opened 2016-06-17
Fix brittle server configuration spec

lebedev-yury:
The `servers_doctor_spec.rb` was randomly failing for me (can be reproduced with `bundle exec rspec --seed 5127`), because the server configuration was not reset after some of the specs.

### DIFF
--- a/spec/lib/capistrano/doctor/servers_doctor_spec.rb
+++ b/spec/lib/capistrano/doctor/servers_doctor_spec.rb
@@ -57,6 +57,7 @@ module Capistrano
       end
 
       it "doesn't fail for no servers" do
+        Capistrano::Configuration.reset!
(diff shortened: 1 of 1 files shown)


lebedev-yury:
I was also thinking on adding the `Capistrano::Configuration.reset!` in `before` block for the whole context, which would add a bit of overhead to the specs, but would make the examples more reliable.

mattbrictson:
Good improvement. Thanks! ✨

mattbrictson:
Oops, I was a bit too fast on the merge button and missed your follow-up comment. If you want to move it to the `before` block, I would accept that as well.

lebedev-yury:
@mattbrictson ok, I will prepare the PR for that, no problem

PR #1707, opened 2016-06-17
Move configuration reset to before block

lebedev-yury:
As was discussed in https://github.com/capistrano/capistrano/pull/1706


mattbrictson:
Nice. 😎

PR #1708, opened 2016-06-17
Add `net-ssh` gem version to `doctor:gems` output

### LINKED ISSUE #1705: Feature: Add net-ssh to list of gems reported by `cap doctor`
Right now the `doctor:gems` task reports gem versions for the following gems:

* capistrano
* airbrussh
* rake 
* sshkit
* *all gems starting with capistrano-*

I think we should add **net-ssh** to this list, because it is a crucial part of Capistrano's behavior, and certain versions of net-ssh have been known to cause bugs (e.g. https://github.com/capistrano/sshkit/issues/357).

If someone wants to try doing a PR for this, here are some pointers:

* Add `net-ssh` here: https://github.com/capistrano/capistrano/blob/b5e4aa6d5e01699960270a9b20bc4b62709cbfb7/lib/capistrano/doctor/gems_doctor.rb#L37
* Add a corresponding test, like this: https://github.com/capistrano/capistrano/blob/b5e4aa6d5e01699960270a9b20bc4b62709cbfb7/spec/lib/capistrano/doctor/gems_doctor_spec.rb#L25-L28
* Update the CHANGELOG

lebedev-yury:
Closes #1705

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * Change git wrapper path to work better with multiple users (@thickpaddy)
   * Make name of current directory configurable via configuration variable `:current_directory` (@websi)
   * `doctor` no longer erroneously warns that `:git_strategy` and other SCM options are "unrecognized" (@shanesaww)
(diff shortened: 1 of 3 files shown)


mattbrictson:
Excellent work. Thanks! 🙌

PR #1711, opened 2016-06-20
Allow validated variables (e.g. :application) to accept proc values

### LINKED ISSUE #1681: Setting :application to a Proc causes cap to crash

mattbrictson:
This fixes a bug where the `:application` validation rule would fail if given a proc value.

The fix involves changing how validation rules are applied to procs. For a proc, instead of validating immediately, we defer validation until when the proc is fetched and called.

Since this causes the validation logic to be non-trivial, I factored it out into a separate class using the decorator pattern. Now `Variables` handles the core set/fetch logic, and `ValidatedVariables` decorates it to layer on validation.

Fixes #1681.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * `doctor` no longer erroneously warns that `:git_strategy` and other SCM options are "unrecognized" (@shanesaww)
   * Add `net-ssh` gem version to `doctor:gems` output (@lebedev-yury)
   * Deprecate `remote_file` feature (will be removed in Capistrano 3.7.0) (@lebedev-yury)
+  * Fix `NoMethodError: undefined method gsub` when setting `:application` to a
+    Proc. [#1681](https://github.com/capistrano/capistrano/issues/1681)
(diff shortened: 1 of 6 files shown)


leehambley:
your implementation of `assert_valid_later` is quite satisfyingly succinct. Quick read-through says LGTM.

mattbrictson:
Thanks! I'll merge this in, then.

deekthesqueak:
I've updated to 3.6.0 as this PR was merged. I have a similar setup to https://github.com/capistrano/capistrano/pull/1683#issuecomment-224347597

This line in my deploy.rb is still causing problems
`set :application, proc { fetch(:stage).split(':').reverse[1] }`

results in 

`NoMethodError: undefined method `gsub' for #<Proc:0x007ffdea180f98@config/deploy.rb:2>`

mattbrictson:
@deekthesqueak You are correct. I will reopen #1681 since this PR is an incomplete fix.

In the meantime, you can use this syntax and it should work:

```ruby
set(:application) { fetch(:stage).split(':').reverse[1] }
```

PR #1714, opened 2016-06-28
Fix latest violations identified by Rubocop 0.41

mattbrictson:
The latest version of Rubocop flags some new violations that were missed previously. This PR fixes them.

* Lint/InheritException: Inherit from RuntimeError instead of Exception.
* Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping. (https://github.com/bbatsov/ruby-style-guide#consistent-string-literals)

### DIFF
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -7,7 +7,7 @@ require_relative "configuration/validated_variables"
 require_relative "configuration/variables"
 
 module Capistrano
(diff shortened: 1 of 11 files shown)


mattbrictson:
@leehambley I'd like to merge this ASAP because it otherwise all Travis builds will start failing due to the new RuboCop gem.

PR #1717, opened 2016-06-30
Skip RuboCop in Travis for Rubinius

mattbrictson:
For some reason Rubinius seg faults sometimes when running RuboCop on Travis. Rather than have periodically failing builds, just disable it. We still get adequate coverage because RuboCop results shouldn't be different for rbx anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,11 @@ rvm:
   - 2.2
   - 2.1
   - 2.0
-  - rbx-2
+matrix:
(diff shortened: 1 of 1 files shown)


will-in-wi:
👍

mattbrictson:
@will-in-wi thanks for the review

will-in-wi:
My pleasure. Thanks for the work you do on this!

PR #1721, opened 2016-07-11
Deprecate :git_strategy, etc. variables and prepare for 3.6.0 release

mattbrictson:
As discussed in #1572, the `:git_strategy`, `:hg_strategy`, and `:svn_strategy` variables will be removed when that PR is merged.

I would like to give users a heads-up by showing deprecation warnings. This PR adds deprecation warnings that are displayed whenever one of these variables is `set`.

My intention is for this PR to be merged and released as Capistrano 3.6.0.

We would then follow up with a 3.7.0 release that includes #1572, which means finally getting the new SCM plugin system out the door.

I also added an UPGRADING-3.7 guide to further explain what is coming with that release.

@leehambley I can manage the 3.6.0 release this week if you agree that it is ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * Fix `NoMethodError: undefined method gsub` when setting `:application` to a
     Proc. [#1681](https://github.com/capistrano/capistrano/issues/1681)
     (@mattbrictson)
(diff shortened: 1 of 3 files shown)


will-in-wi:
👍 per the code change and documentation. I'm excited to see this go out.

mattbrictson:
@leehambley Do you agree on making 3.6.0 a deprecation release with 3.7.0 being the actual SCM refactor?

leehambley:
Yes I do @mattbrictson thanks for the bump

mattbrictson:
OK, I'll merge this in. Thanks for the review!

Issue #1686, opened 2016-05-19
Re-consider how calling already invoked tasks should behave


kenn:
May be a proposal for v4.

Right now, the same task can get called only once, unless `Rake::Task[].reenable` is called explicitly.

``` ruby
namespace :deploy do
  task :restart_unicorn_twice? do
    invoke 'deploy:restart_unicorn'
    invoke 'deploy:restart_unicorn'
  end

  task :restart_unicorn_twice! do
    invoke 'deploy:restart_unicorn'
    Rake::Task['deploy:restart_unicorn'].reenable
    invoke 'deploy:restart_unicorn'
  end
end
```

``` ruby
$ cap staging deploy:restart_unicorn_twice?
00:00 deploy:restart_unicorn
      01 /etc/init.d/unicorn restart
      01 Restarting
    ✔ 01 root@12.34.56.78 6.070s

$ cap staging deploy:restart_unicorn_twice!
00:00 deploy:restart_unicorn
      01 /etc/init.d/unicorn restart
      01 Restarting
    ✔ 01 root@12.34.56.78 6.074s
      01 Restarting
    ✔ 01 root@12.34.56.78 5.936s
```

However, I believe `reenable` should be set for each task by default, because:
- I know that's how rake is designed to work, but it's an implementation detail that users of capistrano shouldn't care. Let's keep in mind that even non-ruby devs are using capistrano for deployment.
- "Run only once" semantics is useful for build tasks with dependencies, which is what rake is originally designed for, but it's almost always a surprising, unwanted behavior for deployment tools. We don't think in terms of dependency. We think in terms of execution order.
- If we DO believe that "run only once" is the right way, we should at least offer a syntax sugar for `Rake::Task[].reenable` - that's very ugly and distracting.

leehambley:
> May be a proposal for v4.

I think it would be safe enough to put that into a "feature" release, we adhere to semver, so bumping from 3-4 will mean we've done a "rewrite".

All valid points, thanks for the contribution, if you wanted to take a shot at a patch we'd be glad to consider it. 

For whatever it's worth we could try an intermediate release that hooks the `Rake::Task[].........` family of functions and directs people here to join the discussion, and we can get a few more people's input?

kenn:
I'm wondering if there's any people who actually wrote a task that depends on the current behavior. For them, it's going to be a breaking change - but it's also believable that there are no one.

Simplest solution would be:

```ruby
module Capistrano::DSL
  def invoke(task, *args)
[...]

PR #1724, opened 2016-07-14
Print a warning if a task is skipped because of re-invocation.

### LINKED ISSUE #1689: Feature: Emit warning when invoke is used on a task that has already been invoked
See discussion in #1686 .

I think we have to monkey patch https://github.com/ruby/rake/blob/master/lib/rake/task.rb#L184 and hook in there (actually given that this only *reads* the value of `@already_invoked` we could just alias chain this method and emit a warning before hand if `@already_invoked` is truthy.

troelskn:
Resolves #1689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
+  * Added warning about future deprecation of reinvocation behaviour (@troelskn)
   * Added a `doctor:servers` subtask that outputs a summary of servers, roles & properties (@irvingwashington)
   * Raise a better error when an ‘after’ hook isn’t found (@jdelStrother)
   * Restrict the uploaded git wrapper script permissions to 700 (@irvingwashington)
--- a/lib/capistrano/all.rb
+++ b/lib/capistrano/all.rb
@@ -11,6 +11,7 @@ require "capistrano/i18n"
 require "capistrano/dsl"
 require "capistrano/application"
 require "capistrano/configuration"
+require "capistrano/rake_monkey_patch"
 
 module Capistrano
 end
--- a/lib/capistrano/rake_monkey_patch.rb
+++ b/lib/capistrano/rake_monkey_patch.rb
@@ -1,0 +1,16 @@
+require "rake/task"
+module Rake
+  class Task
+    alias original_invoke_with_call_chain invoke_with_call_chain # :nodoc:
+    def invoke_with_call_chain(task_args, invocation_chain) # :nodoc:
+      if self.class.ancestors.include?(Capistrano::TaskEnhancements) && already_invoked
+        file, line, = caller[2].split(":")
+        colors = SSHKit::Color.new($stderr)
+        $stderr.puts colors.colorize("Warning: Reinvoking `#{name}' from #{file}:#{line} - Task will be silently skipped.", :red)
+        $stderr.puts colors.colorize("This behaviour is deprecated and will change in a future release.", :red)
+        $stderr.puts colors.colorize("If this affects you, please come to this URL to discuss: https://github.com/capistrano/capistrano/issues/1686", :yellow)
+      end
+      original_invoke_with_call_chain(task_args, invocation_chain)
+    end
+  end
+end
--- a/spec/lib/rake/task_spec.rb
+++ b/spec/lib/rake/task_spec.rb
@@ -1,0 +1,26 @@
+require "spec_helper"
+require "rake"
+require "capistrano/dsl"
+require "capistrano/rake_monkey_patch"
+
+module Rake
+  describe Task do
+    before do
+      Rake::Task.__send__ :include, Capistrano::TaskEnhancements
+    end
+
+    after do
+      # Ensure that any tasks we create in these tests don't pollute other tests
+      Rake::Task.clear
+    end
+
+    it "reinvoking a task will print a message on stderr" do
+      Rake::Task.define_task("some_task")
+
+      Rake::Task["some_task"].invoke
+      expect do
+        Rake::Task["some_task"].invoke
+      end.to output(/^.*Warning: Reinvoking `some_task' from .* - Task will be silently skipped/).to_stderr
+    end
+  end
+end


leehambley:
Please check the build, this seems to break most interpreters.

troelskn:
Yep - Forgot to run rubocop before opening PR. They pass now.

leehambley:
Thanks, it's a shame Rubocop *breaks the build*, it would be nice if Travis could handle more than one task and report tests: OK, lint: broken. 

Thanks!

leehambley:
Can you check the contributing guidelines and credit yourself in the readme, otherwise looks good to merge.

mattbrictson:
I am concerned about the use of `Airbrussh::Colors` here. Two problems:

1. It tightly couples the code to the airbrussh gem. Right now Airbrussh is fairly isolated as a plugin; this would start making it more intertwined.
2. The `Color` module functions do not honor TTY and environment variables; i.e. there is no non-color fallback in consoles that don't support it.

I know the colors in the warning are a nice touch, but for now I would just remove them.

mattbrictson:
Alternatively, you could use `SSHKit::Color`, which addresses both issues.

mattbrictson:
How hard would it be to add a test for this?

troelskn:
@mattbrictson understodd about colours - I had mixed feelings about it too, but I think they add value in that the whole premise of the feature is to provide documentation for newbies. I'll switch to use sshkit's implementation as suggested. It's still a dependency, but I suppose Capistrano is already so heavily intertwined with sshkit that it doesn't matter much from a practical perspective.

I wanted to add a test, but I really have no idea how to do that. Any thoughts on that?  I have some sample code I used for manual testing, to trigger the feature, that could for the base of an automated test.

mattbrictson:
For writing a test, there are two ways I would try to do it.

First, you could do a feature test. This is an end-to-end test that actually runs Capistrano against a VM. The drawback is that if you don't have your environment set up already (Virtual Box, etc.), it may take some time to get started. (See DEVELOPMENT.md for instructions.) Also if you've never used Cucumber before, there is a learning curve. Take a look at the existing tests in `features/*.feature` to get an idea.

Alternatively, you could write a spec that calls the DSL directly and verifies the output. This may be more straightforward. A good template to follow would be `servers_doctor_spec.rb`. You can see how it uses `include Capistrano::DSL` to get access to the DSL methods (i.e. so you can call `invoke`), and `expect { ... } to output("...").to_stderr` for testing output. You may also need to `include Rake::DSL` to be able to define Rake tasks in your test. Don't forget to call `Rake::Task.clear` in an `after` block to reset Rake to a clean slate.

troelskn:
That fixes the colour issue. I'll have a go at the test case.

troelskn:
Added `spec/lib/rake/task_spec.rb`.

troelskn:
OK. Ready for another review.

> Can you check the contributing guidelines and credit yourself in the readme, otherwise looks good to merge.

I don't see any list of contributors?

mattbrictson:
> I don't see any list of contributors?

You can take credit by adding a CHANGELOG entry and putting your @ GitHub username at the end of the entry. Totally optional.

mattbrictson:
Looks good to merge!

@troelskn I'll wait to hear whether you want to add a CHANGELOG entry.

troelskn:
Added an entry to changelog - guess it's important enough, since it's announcing a possible future change of behaviour.

mattbrictson:
Thanks! 🙇

mattbrictson:
Unfortunately I'm going to have to revert this PR based on further real world testing. During a typical deploy using capistrano-rails, it outputs several warnings for the `deploy:set_rails_env` task:

```
Warning: Reinvoking `deploy:set_rails_env' from gems/rake-11.2.2/lib/rake/task.rb:207 - Task will be silently skipped.
This behaviour is deprecated and will change in a future release.
```

I think this is because `deploy:set_rails_env` is a *prerequisite* of many capistrano-rails tasks. So even though we aren't explicitly using `invoke` on it, Rake is implicitly invoking the task every time a task further down the dependency chain is run.

You can also see that the deprecation warning is pointing to a line number inside the Rake gem itself.

So, I think we need to approach this differently. Rather than monkey patching `Rake::Task`, maybe make a change to `Capistrano::DSL#invoke` instead.

troelskn:
I see your point.

Problem is that this muddies the semantics somewhat. If *some* tasks are idempotent (and thus shouldn't be re-invoked), but others should be re-invoked, how do you know the difference? At the very least, I think it should be the task itself that specifies whether it's idempotent or not - It shouldn't be the calling context that determines it.

I haven't thought this through, but could we somehow set an option on a task to specify which semantics it follows? Should we call this something other than "task" to distinguish the two? Am I totally overthinking this?

troelskn:
How about something like this:

```
require "rake/task"
module Rake
  class Task
    def reinvokable!
      @reinvokable = true
    end

    alias original_invoke_with_call_chain invoke_with_call_chain # :nodoc:
    def invoke_with_call_chain(task_args, invocation_chain) # :nodoc:
      original_invoke_with_call_chain(task_args, invocation_chain)
    ensure
      if @reinvokable
        @already_invoked = false
      end
    end
  end
end
```

Usage:

```
desc "A good task"
task :good do
  puts "yah"
end.reinvokable!

desc "calling twice"
task :twice do
  invoke "good"
  invoke "good"
end
```

And coupled with a warning from `Capistrano::DSL#invoke` if a non-reinvokable is invoked twice.

leehambley:
I do like the idea, I wonder if it could rather be an argument to the `task` DSL function though. 

Mixed feelings, I don't like the monkey patch (but could live with it) or the syntax (but it's a private API, so I could live with it)

The resulting functionality seems to align with sanity.

Alternative is to whitelist "tasks" which we know are prerequisites.

Finally we might add `prerequisite_task` to the DSL (most preferred) and use that for our "only called as a prerequisite" tasks.

What do you think?

troelskn:
Agree that the syntax is weird. Task arguments are already syntactically quite overloaded for rake/capistrano, but I imagine it's possible to wedge it in there somehow. Another option is (like your suggestion with `prerequisite_task `) to make a new dsl-helper method that hides the uglyness away. In any case, this is just syntax.

> Alternative is to whitelist "tasks" which we know are prerequisites.
> ...
> Finally we might add prerequisite_task to the DSL (most preferred) and use that for our "only called as a prerequisite" tasks.

Except for the syntax, those two are same, right? What you're talking about is to change the default semantics of tasks, but retain the rake-semantics in a new construct (The pre-requisite task). I know this is where we started off, but my spider sense tells me that it would be better to stay as close to original Rake semantics as possible?

leehambley:
Yeh, we're only talking about syntax. In the end I think we've got one `task` DSL method, but two distinct kinds of tasks... 

I'm thinking that a `prerequisite` method that allows us to define a task that's allowed to be invoked once, and only once, and we change the meaning of the `task` to specifically be something that we can invoke as many times as we like.

That's quite a wide departure from Rake, but I don't think that's a bad thing.

Matt tends to have a kind of insight and clarity on these things that I can't compete with :) Let's wait for him to weigh-in.

mattbrictson:
I think where we left the discussion in #1686 is that we're not sure yet how much we want to change the Rake execution model. For now, we resolved to display a warning, so that users that call `invoke` are not caught off guard when nothing happens.

I suggest changing `Capistrano::DSL#invoke` so that it:

1. Checks that the named task exists
2. Checks `Task#already_invoked`
3. Prints a warning if the task was already invoked
4. Calls `Task#invoke` regardless

The warning could say:

> Skipping #{task}.
Capistrano tasks may only be invoked once. Since task "#{task}" was previously invoked, `invoke("#{task}")` on #{file}##{line} will be skipped. To force this task to run again, first call `Rake::Task["#{task}"].reenable`. THIS BEHAVIOR MAY CHANGE IN A FUTURE VERSION OF CAPISTRANO. Please join the conversation here if this affects you.
`https://github.com/capistrano/capistrano/issues/1686`

troelskn:
Good point - we're getting ahead of our selves here. I'll augment the patch to modify `Capistrano::DSL#invoke` instead. We can then take some time to consider if we want rake-behaviour per default or as opt-in.

Issue #1728, opened 2016-07-19
NoMethodError: undefined method `validate' for main:Object


zhanghqgit:
cap version :3.4.0

Gemfile:
```
gem 'capistrano', '3.4.0'
gem 'capistrano-daemonize'
gem 'highline'
```

Capfile:
```
# Load DSL and set up stages
require 'capistrano/setup'

# Include default deployment tasks
require 'capistrano/deploy'

# Include tasks from other gems included in your Gemfile
#
# For documentation on these, see for example:
#
#   https://github.com/capistrano/rvm
#   https://github.com/capistrano/rbenv
#   https://github.com/capistrano/chruby
#   https://github.com/capistrano/bundler
#   https://github.com/capistrano/rails
#   https://github.com/capistrano/passenger
#
# require 'capistrano/rvm'
# require 'capistrano/rbenv'
# require 'capistrano/chruby'
# require 'capistrano/bundler'
# require 'capistrano/rails/assets'
# require 'capistrano/rails/migrations'
# require 'capistrano/passenger'

# Load custom tasks from `lib/capistrano/tasks` if you have any defined
Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
````

then in  deploy.rb, i define a validation like below:
```
validate :project_to_deploy do |key,value|
  .....
end

```

mattbrictson:
Hi, the `validate` feature is not available in Capistrano 3.4.0. You need to upgrade to Capistrano 3.5.0.

The changelog is incorrect in this case. (It says the validation feature was introduced in 3.3.3, which is wrong.) This mistake was probably due to a bad merge. I'll fix it.

zhanghqgit:
thanks a lot

zhanghqgit:
also for the Capistrano::ValidationError

PR #1729, opened 2016-07-19
Move validation release notes to 3.5.0

mattbrictson:
The release notes for the validation feature were incorrectly placed in the 3.3.5 section of the changelog for some reason. I double-checked the Git history and confirmed that the validation code was not actually released until 3.5.0.

This was brought to my attention in #1728.


leehambley:
Wow, I wonder how that happened, seems rather "precise" to have been a bad merge as you suggested. Good catch, thanks for the patch.

PR #1730, opened 2016-07-19
Print a warning if a task is skipped because of re-invocation.

### LINKED ISSUE #1689: Feature: Emit warning when invoke is used on a task that has already been invoked

troelskn:
Resolves #1689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
 
+  * Added warning about future deprecation of reinvocation behaviour (@troelskn)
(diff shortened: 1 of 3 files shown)


troelskn:
As per discussion here: https://github.com/capistrano/capistrano/pull/1724#issuecomment-233398023

mattbrictson:
Very nice, thanks! I tested this against one of my own projects and this warning actually exposed a bug where I was inadvertently calling `invoke` twice. Works as advertised. 👍

PR #1731, opened 2016-07-20
Specify directory name for rollback. Resolves #1155

### LINKED ISSUE #1155: Be able to specify directory name for deployment and rollback
There are some cases in distributed developments that a number of servers become temporarily unreachable. Most of the time, this can be solved by redeploying but it gets a bit tricky when you try to rollback. The way rollback works at the time, from what I understand, is to map the symlink to the second to the last directory (which works because directories are named with timestamps). In cases like I mentioned, a real rollback will not work because the servers will have different "second to the last directory". I understand that creating a custom deploy/rollback task would probably solve the issue but is there any discussion regarding this to have it as a core feature?

```
  # rollback
  task :rollback_release_path do
    on release_roles(:all) do
      releases = capture(:ls, '-xr', releases_path).split
      if releases.count < 2
        error t(:cannot_rollback)
        exit 1
      end
      last_release = releases[1]
      set_release_path(last_release)
      set(:rollback_timestamp, last_release)
    end
  end
```

lanrion:
Resolves #1155

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ https://github.com/capistrano/capistrano/compare/v3.5.0...HEAD
   * Deprecate `:git_strategy`, `:hg_strategy`, and `:svn_strategy` variables.
     These will be completely removed in 3.7.0. Refer to the
     [upgrading doc](UPGRADING-3.7.md) for details.
+  * It is now possible to rollback to a specific release using the
+    `ROLLBACK_RELEASE` environment variable.
+    [#1155](https://github.com/capistrano/capistrano/issues/1155) (@lanrion)
 
 ## `3.5.0`
 
--- a/lib/capistrano/i18n.rb
+++ b/lib/capistrano/i18n.rb
@@ -15,6 +15,7 @@ en = {
   no_old_releases: "No old releases (keeping newest %{keep_releases}) on %{host}",
   linked_file_does_not_exist: "linked file %{file} does not exist on %{host}",
   cannot_rollback: "There are no older releases to rollback to",
+  cannot_found_rollback_release: "Cannot rollback because release %{release} does not exist",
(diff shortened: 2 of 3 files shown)


mattbrictson:
Hi, thanks very much for the PR and for including a changelog entry.

I think in general the implementation looks fine, but I would like to suggest some changes:

* `V` is an extremely terse environment variable name. How about `ROLLBACK_RELEASE` instead?
* If someone specifies a release via the environment variable, and that release is not found in the list of releases, then your code shows the `:cannot_rollback` error and exits. However that message says "There are no older releases to rollback to", which is misleading. Shouldn't the error in this case say something like "Cannot rollback because release ###### does not exist"?
* Can you expand upon the changelog entry to explain the new feature? Something like: "It is now possible to rollback to a specific release using the ROLLBACK_RELEASE environment variable."
* Since this is a brand new feature, can you please also open a PR in https://github.com/capistrano/documentation to add documentation for it on the [rollbacks](http://capistranorb.com/documentation/getting-started/rollbacks/) page?

Finally: Capistrano currently does not have any test coverage for the rollback process. Normally I would not accept a PR without tests, but since there is no precedent for rollback tests I think it is unfair to require you add it for this PR. However if you have any suggestions  for how to test this, that would also be appreciated.

lanrion:
Thank you for your suggestions, I updated the commit and opened a PR in https://github.com/capistrano/documentation/pull/195 on the rollbacks page.

lanrion:
I have no idea about test rollback task with Rspec,  I test it in my real project manually。Maybe we can make a simulation to deploy on CI?

mattbrictson:
I think this looks good as is.

@leehambley any objections?

mattbrictson:
Thanks for the great PR! ✨

PR #1733, opened 2016-07-21
Update version number in README

mattbrictson:
Update version number in README to the current one (3.5). Also add a step to the RELEASING document so that we are less likely to forget this in the future.


leehambley:
Good call on `RELEASING`

PR #1739, opened 2016-07-28
Ensure validation works for callable values

### LINKED ISSUE #1681: Setting :application to a Proc causes cap to crash

mattbrictson:
Fixes #1681, again!

The previous solution worked for validating values set using the block syntax, like this: `set(:application) { "value" }`. However it failed for the much more common scenario of `set :application, -> { "value" }`.

To fix this, we have to check whether the value passed to `set` is callable (i.e. responds to `call` with no arguments). If so, we apply the deferred validation logic in this scenario as well.

The additional wrinkle is the case where the value is a `Question` object. A Question responds to `call`, but it is not a normal Ruby Proc. We can't wrap it in another object to add validation because then `Configuration#is_question?` (which uses `is_a?(Question)`) would erroneously return false.

To account for this, I have changed the technique used for deferred validation to use `define_singleton_method` to wrap just the `call` method instead of wrapping the entire object. This ensures that `is_a?(Question)` remains true even after validation is applied.

Finally, in an effort to keep backtraces as clean as possible, all of this validation wrapping stuff is only applied if a validation rule is actually registered. If no validation rules are defined for a given key, there is no benefit to this meta-programming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.6.0...HEAD
 
+* Fix `NoMethodError: undefined method gsub` when setting `:application` to a Proc, for real this time! The original fix released in 3.6.0 worked for values specified with blocks, but not for those specified with procs or lambdas (the latter syntax is much more common). [#1681](https://github.com/capistrano/capistrano/issues/1681)
 * Your contribution here!
 
 ## `3.6.0` (2016-07-26)
--- a/lib/capistrano/configuration/validated_variables.rb
+++ b/lib/capistrano/configuration/validated_variables.rb
@@ -7,9 +7,9 @@ module Capistrano
     # user-supplied validation rules. Each rule for a given key is invoked
     # immediately whenever `set` is called with a value for that key.
     #
-    # If `set` is called with a block, validation is not performed immediately.
-    # Instead, the validation rules are invoked the first time `fetch` is used
-    # to access the value.
+    # If `set` is called with a callable value or a block, validation is not
+    # performed immediately. Instead, the validation rules are invoked the first
+    # time `fetch` is used to access the value.
     #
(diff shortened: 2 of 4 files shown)


mattbrictson:
@leehambley I would appreciate a second set of eyes on this PR, if you have a chance. Meta-programming is almost always too clever for its own good, but it seemed the least-bad option. Maybe I missed a simpler solution.

mattbrictson:
I don't want to put anyone in particular on the spot, but if someone passing by has time to look at this PR, I would appreciate it! I don't feel comfortable merging this without a code review.

will-in-wi:
This is clever…

I'm struggling to think of a better approach. Patching `#call` feels dirty to me, but the only reasonable alternative which retains typing would be to dynamically create a decorator class which inherits from the class in question.

Alternatively, could we solve the `Question` case by not type-checking `is_a? Question` but rather check for some relevant method or something?

If not, I think this implementation looks solid. I appreciate the tests and the comments, it would be tricky to grok what is going on and why without it.

Thanks!

will-in-wi:
I suppose a terrible hack would be to create a decorator class that forwards `#is_a?` and `#kind_of?` to the composed object…

will-in-wi:
Could we approach it by taking the callable and wrapping it in another callable which handles validation? Then we could just handle the `Question` as a special case with a `ValidatedQuestion < Question` class…

Thinking out loud.

mattbrictson:
> we could just handle the Question as a special case with a `ValidatedQuestion < Question` class…

Yes, I had the same thought. So that would mean:

* For the non-Question case, we simply wrap the original proc in another proc
* For the Question case, create a new `ValidatedQuestion` that has a `call` method that does the same thing

I will give that a shot.

mattbrictson:
Closing in favor of #1744.

Issue #1738, opened 2016-07-27
Git SCM fails when :local_user contains a space


fnordfish:
#### Steps to reproduce

1. Use the GIT SCM strategy
2. Setup the `:local_user` variable so that it contains a full name (eg: `Robert Schulze`. actualy, I was reading it from `ENV["GIT_COMMITTER_NAME"] || ENV["GIT_AUTHOR_NAME"]`)
3. Try to deploy

#### Expected behaviour

The `git-ssh` wrapper script should be callable.

#### Actual behaviour

```
 DEBUG Uploading /tmp/git-ssh-project-staging-Robert Schulze.sh 0.0%
  INFO Uploading /tmp/git-ssh-project-staging-Robert Schulze.sh 100.0%
  INFO [8bce1bd1] Running /usr/bin/env chmod 700 /tmp/git-ssh-project-staging-Robert Schulze.sh as user@host
...
DEBUG [8bce1bd1] 	chmod:
 DEBUG [8bce1bd1] 	cannot access ‘/tmp/git-ssh-project-staging-Robert’
 DEBUG [8bce1bd1] 	: No such file or directory
 DEBUG [8bce1bd1]
 DEBUG [8bce1bd1] 	chmod:
 DEBUG [8bce1bd1] 	cannot access ‘Schulze.sh’
 DEBUG [8bce1bd1] 	: No such file or directory
 DEBUG [8bce1bd1]
```

Looking at the target host, I can see, that the file is correctly created: `/tmp/git-ssh-project-staging-Robert Schulze.sh`.

mattbrictson:
Thanks for the detailed report! I agree, this is a bug. Would you be comfortable contributing a PR? I think a `gsub` call to replace non-alphanumerics with hyphens when constructing the `:git_wrapper_path` would be a reasonable solution.

https://github.com/capistrano/capistrano/blob/b751b68b8d99c05236f744be5da49e42ddff8c06/lib/capistrano/tasks/git.rake#L9-L10

fnordfish:
Sorry, but I actually won't be able to make the time soon. So, this one is up for grabs :)

mattbrictson:
OK, first come first served, whoever wants to take it! 😄

will-in-wi:
The fix itself is trivial, but I'm not sure where to write a test for it. Any suggestions?

Also, is this still relevant after https://github.com/capistrano/capistrano/pull/1572 goes in?

mattbrictson:
[...]

PR #1742, opened 2016-07-30
Fix space in git suffix. Fixes 1738

will-in-wi:
Fix for https://github.com/capistrano/capistrano/issues/1738

Still trying to figure out a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.6.0...HEAD
 
 * Your contribution here!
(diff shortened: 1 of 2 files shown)


mattbrictson:
Thanks for the PR! Testing might be too difficult to accomplish given the current codebase. Feel free to skip the test this time around. It will be easier to add a test once #1572 is merged.

If you could add a changelog entry I will merge this in.

mattbrictson:
Although perhaps `gsub(/\s+/, " ")` would be better since it would catch all sorts of white space?

will-in-wi:
I added the changelog entry and swapped `#tr` for `#gsub`.

mattbrictson:
Looks good. Thanks! ✨

PR #1744, opened 2016-08-03
Ensure validation works for callable values

### LINKED ISSUE #1681: Setting :application to a Proc causes cap to crash

mattbrictson:
Fixes #1681. Third time's the charm!

The original code worked for validating values set using the block syntax, like this: `set(:application) { "value" }`. However it failed for the much more common scenario of `set :application, -> { "value" }`.

To fix this, we have to check whether the value passed to set is callable (i.e. responds to `call` with no arguments). If so, we apply the deferred validation logic in this scenario as well.

The additional wrinkle is the case where the value is a `Question` object. A Question responds to `call`, but it is not a normal Ruby Proc. We can't simply wrap it in another Proc to add validation because then `Configuration#is_question?` would erroneously return false.

To account for this, I use a trivial `ValidatedQuestion` class that simply delegates to the Proc that does the validation. This class is a subclass of `Question`, so the `Configuration#is_question?` test still works.

Finally, in an effort to keep backtraces as clean as possible, all of this validation wrapping stuff is only applied if a validation rule is actually registered. If no validation rules are defined for a given key,
there is no benefit to this meta-programming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Reverse Chronological Order:
 
 https://github.com/capistrano/capistrano/compare/v3.6.0...HEAD
 
+* Fix `NoMethodError: undefined method gsub` when setting `:application` to a Proc, for real this time! The original fix released in 3.6.0 worked for values specified with blocks, but not for those specified with procs or lambdas (the latter syntax is much more common). [#1681](https://github.com/capistrano/capistrano/issues/1681)
 * Your contribution here!
 
 ## `3.6.0` (2016-07-26)
--- a/lib/capistrano/configuration/validated_variables.rb
+++ b/lib/capistrano/configuration/validated_variables.rb
@@ -7,9 +7,9 @@ module Capistrano
     # user-supplied validation rules. Each rule for a given key is invoked
     # immediately whenever `set` is called with a value for that key.
     #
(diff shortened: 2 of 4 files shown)


mattbrictson:
@will-in-wi How's this look? I'm using the `class ValidatedQuestion < Question` trick you suggested. Now there is no meta-programming.

will-in-wi:
I like this solution better, I think.

The long term solution is probably to figure out how to make `Configuration#is_question?` not type check, as this limits the ability of someone to create a new question type object. I'm not sure what it is actually used for, but if my initial guess is right, the main reason to want to know if something is a question or not is to determine whether it is about to request user input. It might be better to change the definition of Question to be an object which responds to the method `#requests_input?` with true. If it doesn't respond, or returns false, it isn't a question. However, I'm not sure I understand the purpose of Question, so I might be totally off base. All of the deploys I've created have had to be fully automatable, so asking for input hasn't been allowed.

Thanks for working on this!

mattbrictson:
> The long term solution is probably to figure out how to make Configuration#is_question? not type check

I agree. I'll open an issue for that.

Thanks for the review!

PR #1767, opened 2016-08-31
Document new SCM plugin system

mattbrictson:
This is the corresponding documentation for the new SCM plugin system implemented here: https://github.com/capistrano/capistrano/pull/1572


mattbrictson:
Merging now that #1572 is merged.

PR #1768, opened 2016-08-31
Switch from redcarpet to GH-flavored kramdown

### LINKED ISSUE #1761: GitHub Pages emits warning about Markdown engine
_From @mattbrictson on February 9, 2016 1:5_

:weary: 

> You are currently using the 'redcarpet' Markdown engine, which will not be supported on GitHub Pages after May 1st. At that time, your site will use 'kramdown' for markdown rendering instead. To suppress this warning, remove the 'markdown' setting in your site's '_config.yml' file and confirm your site renders as expected. For more information, see https://help.github.com/articles/updating-your-markdown-processor-to-kramdown.

_Copied from original issue: capistrano/documentation#172_

mattbrictson:
This should fix #1761, but it has a side effect of completely changing the syntax highlighting.

Before:
screen shot 2016-02-12 at 2 36 10 pm

After:
screen shot 2016-02-12 at 2 36 23 pm

@leehambley Do you remember enough about the CSS to troubleshoot this? It looks like the syntax highlight classes are completely different, so I don't think it is a quick fix.

At minimum, we should:
- [ ] Add a background color to the code blocks to delineate them
- [ ] Restore the fixed width and overflow: scroll behavior


will-in-wi:
I took a shot at cleaning up styles. Instead of porting the CSS, I just grabbed a similar theme.

It looks like Kramdown uses Pygments compatible classes. And since our existing theme is based on monokai, I just grabbed that theme directly.

https://github.com/will-in-wi/capistrano/commit/9339a63adebdd62ba7d6a7cfc2b995d1827ce777

mattbrictson:
@will-in-wi Thanks, that helps. I pulled in your commit. It looks like we still need to fix the background color issue. If you're able to help with that too, that would be much appreciated. 😃

olleolleolle:
@mattbrictson I amended the monokai.css using this at the top, would this complete the background-color fix and the overflow?

```css
.highlighter-rouge {
 background-color: #272822;
}

.highlight {
 background: #272822;
 padding: 1em;
 font-size: 0.9em;
 overflow: auto;
 line-height: 150%;
 overflow-y: auto;
 overflow-x: scroll;
}
```

(Added 150% line-height to regain visual look from before.)

screenshot 2017-02-04 17 57 26

leehambley:
Looks good to me, thanks for including a screenshot

Am 04.02.2017 17:58 schrieb "Olle Jonsson" <notifications@github.com>:

mattbrictson:
Thanks @olleolleolle, looks great! I've incorporated your change and will merge this in.

Issue #1766, opened 2016-08-29
Host filter doesn't exactly match


cseeger-epages:
Greedy host filter
---

#### Steps to reproduce
eg. cap stage --hosts=system01 <some task>
using a stage with defined hosts:
system011
system012
system020

#### Expected behaviour

empty host list or ERROR Host does not exists or sth. like this

#### Actual behaviour

matching any host starting with the hosts parameter

in this case
system011 and system012

#### Fix

from: https://github.com/capistrano/capistrano/blob/master/lib/capistrano/configuration/host_filter.rb#L17-L27
```ruby
def regex_matcher(values)
        values.map! do |v|
          case v
          when Regexp then v
          else
            vs = v.to_s
            vs =~ /^[-A-Za-z0-9.]+$/ ? vs : Regexp.new(vs)
          end
        end
        Regexp.union values
      end
```

Line 23 needs to be changed from
```ruby
vs =~ /^[-A-Za-z0-9.]+$/ ? vs : Regexp.new(vs)
```
to
```ruby
vs =~ /^[-A-Za-z0-9.]+$/ ? %r{^#{vs}$} : Regexp.new(vs)
```

#### Addition

Wildcard matching should be used while directly using a regex in capistrano 

eg. cap stage --hosts='system01\d' <some task>

using only a string it should match exactly

mattbrictson:
Yes, I believe this is a bug. The [documentation](http://capistranorb.com/documentation/advanced-features/host-filtering/) implies that the strings should be matched exactly, unless certain regular expression syntax is used.

@leehambley @will-in-wi @cshaffer Do you agree this is a bug?

---

Also, I don't think `%r{^#{vs}$}` is exactly correct, since a `.` in the filter string will be interpreted as a matching any single character. Probably what we want is `%r{^#{Regexp.quote(vs)}$}`?

will-in-wi:
I agree that this appears to be a bug. Unless @cseeger-epages is interested in doing so, I can probably whip up a PR.

mattbrictson:
@will-in-wi Thanks. Can you take the lead on seeing this issue is ultimately resolved? I'll assign it to you if that's OK.

will-in-wi:
Sounds good to me. @cseeger-epages: Are you interested in making a pull request for this? You already found the offending line. The only additional thing I would suggest would be to make an rspec test that replicates this issue, and another one which checks for the case when what is passed contains a dot.

cseeger-epages:
[...]

PR #1769, opened 2016-08-31
Update host_filter.rb

cseeger-epages:
possible fix for #1766

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ https://github.com/capistrano/capistrano/compare/v3.6.1...HEAD
 
 * Your contribution here!
 * `remote_file` feature has been removed and is no longer available to use @SaiVardhan
+* fix bug with greedy matching in host_filter (@cseeger-epages)
 
 ## `3.6.1` (2016-08-23)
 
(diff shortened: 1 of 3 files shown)


will-in-wi:
The test failure is Rubocop carping about using `/regex/` instead of `%r{regex}` in this particular case. You should be able to run `bundle exec rubocop -a` and it'll fix the issue for you.

Could you add a couple tests for this change? There are existing tests in `spec/lib/capistrano/configuration/role_filter_spec.rb`. I'd just add one which tries to match `server`. It would return all of them before this fix and none after.

Thanks!

will-in-wi:
Could you also add a changelog entry and give yourself credit?

cseeger-epages:
i will take a look at the tests now

will-in-wi:
Thanks for working on this!

will-in-wi:
👍 I think this looks good. @mattbrictson: merge?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ https://github.com/capistrano/capistrano/compare/v3.6.1...HEAD
 
 * Your contribution here!
 * `remote_file` feature has been removed and is no longer available to use @SaiVardhan
+* fix bug with greedy matching in host_filter (@cseeger-epages)
```
mattbrictson: Could you also reference the issue number? E.g.

```
[#1766](https://github.com/capistrano/capistrano/issues/1766)
```

mattbrictson:
Please see me line comment regarding the changelog. Otherwise this looks good. Thanks!

mattbrictson:
Thank you!

Issue #1748, opened 2016-08-10
within Directory not working


poc7667:
I tried to execute command under this folder `fetch(:latest_release_directory)`
However, I still have to append  `cd #{fetch(:latest_release_directory)} &&` before executing it.
Otherwise I will got exception, because it didn't switch to the target folder at all.

  gem "capistrano",  "~> 3.4.0", require: false

```
on roles(:app) do |host|
  within fetch(:latest_release_directory) do
      execute %Q( cd #{fetch(:latest_release_directory)} && sudo foreman export upstart -e .env /etc/init -a #{ENV['APP_NAME']} -u #{ENV['DEPLOYER']} -l /tmp/#{ENV['APP_NAME']}.log )
  end
end
```

leehambley:
%Q makes a string and disables command mapping. Please see the
documentation for Capistrano and SSHKit, this is thoroughly documented.

(Sent from my Nexus 6, please forgive typos!)

will-in-wi:
@poc7667 Did this fix the issue?

I created this test case: https://gist.github.com/will-in-wi/527327e31af30b3eae2068e2965be05b

From that, I got the following output:

```
$ bundle exec cap production test_case:testing
/tmp
00:00 test_case:testing
      01 pwd
      01 /tmp
    ✔ 01 localhost 2.883s
```

`pwd` and `puts` both returned the correct directory for me.

PR #1770, opened 2016-08-31
Pull request and issue templates for GitHub

will-in-wi:
I gave a first crack at default issue and PR templates.

Thoughts?


will-in-wi:
I should give credit. This is based on the Rails templates.

mattbrictson:
Thanks!

We already have a `./issue_template.md` that @leehambley set up, I believe; can you merge your suggestions with that?

I'll have more time to do a full review on Friday.

will-in-wi:
Will do. I'll try and get the improvement done before Friday.

leehambley:
@will-in-wi I believe the only part of mine that I prefer over yours is the part about:

> Important: GitHub issues are for feature requests or bug reports. The Capistrano team recommends you use Stack Overflow for general questions. For more details, please see our contribution policy.

```
**Important:** GitHub issues are for feature requests or bug reports. The Capistrano team recommends you use [Stack Overflow](http://stackoverflow.com/questions/tagged/capistrano) for general questions. For more details, please see our [contribution policy](https://github.com/capistrano/capistrano/blob/master/CONTRIBUTING.md).
```

Otherwise looks good to me

will-in-wi:
Thanks! I've pushed the changes. Let me know what you think.

.github/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
...
+Provide a general description of the code changes in your pull
+request... were there any bugs you had fixed? If so, mention them. If
+these bugs have open GitHub issues, be sure to tag them here as well,
+to keep the conversation linked together.
+
+### Short checklist
+
+[ ] Did you run `bundle exec rubocop -a` to fix linter issues?
+[ ] If relevant, did you create a test?
+[ ] Did you confirm that the RSpec tests pass?
```
mattbrictson: To get these to appear as actual checkboxes in GitHub, I believe the correct Markdown syntax is:

```
- [ ] 
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Other Information
+
+If there's anything else that's important and relevant to your pull
+request, mention that information here.
+
+If you are fixing a bug or introducing a new feature, please add a CHANGELOG entry.
```
mattbrictson: How about moving this to the "Short checklist" section? I feel like the CHANGELOG is the most commonly-overlooked step, so it should be prominent.

.github/issue_template.md
```diff
@@ -0,0 +1,17 @@
+**Important:** GitHub issues are for feature requests or bug reports. The Capistrano team recommends you use [Stack Overflow](http://stackoverflow.com/questions/tagged/capistrano) for general questions. For more details, please see our [contribution policy](https://github.com/capistrano/capistrano/blob/master/CONTRIBUTING.md).
+
+---
+
+### Steps to reproduce
+If at all possible, please try to create a new Capistrano example project which reproduces the issue you are seeing. You can then upload the individual files to a GitHub Gist or a GitHub project. Often times, an issue is resolved simply by making this test case.
```
mattbrictson: I agree this makes the maintainers' lives easier, but I think this is daunting to ask of someone reporting a bug. You can't really create a  working example Capistrano project without real servers and a `:repo_url` that points to real source code, none of which can be accessed without the reporter's credentials. Unless I am misunderstanding what you are asking for here?

mattbrictson:
@will-in-wi Thanks for putting this together. I added some inline comments. My main concern is that we may discourage people from reporting valid issues if we make the issue template too demanding. The PR template looks excellent and I only suggested some minor edits.

.github/issue_template.md
will-in-wi: I understand the concern… I'm mostly thinking of the situation in https://github.com/capistrano/capistrano/issues/1748 where I was able to create https://gist.github.com/will-in-wi/527327e31af30b3eae2068e2965be05b which was the skeleton of a reproducible case.

Perhaps weakening the language would help, e.g. "if at all possible" should probably be "if reasonable" or "if you are running into an issue in Capistrano itself".

Does it make sense what I'm getting at? I'm not sure what the right approach is.

will-in-wi:
Thanks for your comments! I agree with the concern regarding making the template less demanding and have added some inline comments in that vein.

.github/issue_template.md
mattbrictson: I think your example Gist does a great job of illustrating what this paragraph is asking for. Could you reference that Gist with a link as part of that text? That would make this ready to ship, IMO.

will-in-wi: I have now done this.

mattbrictson:
Looks ready to merge! Any last-minute edits?

Issue #1776, opened 2016-09-13
Custom Filters


DylanFrese:
I would like to implement my own custom (command-line) filters, that apply to `on` like `HOSTS` or `ROLES`. An example would be `REGIONS`, or extending `HOSTS` to match SaltStack minion IDs as well as FQDNs/IPs.

However, [filter.rb](https://github.com/capistrano/capistrano/blob/master/lib/capistrano/configuration/filter.rb) and [configuration.rb](https://github.com/capistrano/capistrano/blob/master/lib/capistrano/configuration.rb#L109) are rather 'closed', making it difficult to do this, and I have to monkey-patch in my own filters.

I am aware of using property filtering to restrict individual tasks to certain targets in the deploy, but that is not quite what I'm looking for.

mattbrictson:
My gut reaction is that this is not something that we want to support in Capistrano core. However I understand your frustration about the lack of a clean extension point for you to add custom filtering. 

Perhaps the best way to accommodate this is via a custom task. You could write a rake task called `regions` that accepts an argument (in other words, it can be invoke by `regions[arg]`). That task could go through the list of defined servers and remove the ones the don't match the filter.

```
cap production regions[123] deploy
```

We might need to change Capistrano to provide a method for you to remove a server, but doesn't seem like a big deal.

Would something like that work?

Ultimately if we're going to support this, I would like it to be done with the smallest change possible. I don't see custom filters as a common use case, so the solution would have to be very low risk.

DylanFrese:
I think that even simpler would just to be to open up the list of filters in `configuration.rb` for appending. Monkey-patching or refining if I want to change the behaviour of `HOSTS` or `ROLES` is not too bad.

mattbrictson:
OK, can post a quick patch or open a PR with what you have in mind? We can work from there.

will-in-wi:
I'm a fan of making this possible. I personally use hostnames and the host regex option to accomplish it at present. My use case is that we have several locations, and we prepend a two char location code to our server names. For example, our servers at our main location are hqsomething01, while our servers in our California location are casomething01. So I can limit a deploy by running `cap prod --hosts=^hq deploy`.

But if we didn't have the servers named this way, I'd need more granular deployment capabilities as defined above.

I'll help out as I can.

Incidentally, a hack which can accomplish this is using env variables. For example you can set up your `config/deploy/production.rb` like so:

```ruby
if ENV['REGION_ONLY'] == 'ca'
[...]

PR #1777, opened 2016-09-14
Add support for adding custom filters

DylanFrese:
Users of the API may now add their own 'on'-filters by calling
Configuration#add_filter, e.g:
`Capistrano::Configuration.env.add_filter(RegionFilter.new('NE'))`.

An filter object passed into this method must respond to .filter, and
this method should accept an array of Server objects and return a subset
of this array.

Addresses Issue #1776

### DIFF
--- a/docs/documentation/advanced-features/custom-filters/index.markdown
+++ b/docs/documentation/advanced-features/custom-filters/index.markdown
@@ -1,0 +1,83 @@
+---
+title: Custom Filters
+layout: default
+---
+
+Custom filters (specifically, Custom On-Filters) limit the hosts that are being
+deployed to, in the same way as
+[Host](/documentation/advanced-features/host-filtering/) and
+[Role](/documentation/advanced-features/role-filtering/) filters, but the exact
+method used to filter servers is up to the user of Capistrano.
+
+Filters may be added to Capistrano's list of filters by using the
+`Configuration#add_filter` method.  Filters must respond to a `filter` method,
+which will be given an array of Servers, and should return a subset of that
+array (the servers which passed the filter).
+
+`Configuration#add_filter` may also take a block, in which case the block is
+expected to be unary. The block will be passed an array of servers, and is
+expected to return a subset of that array.
+
+Either a block or object may be passed to `add_filter`, but not both.
+
+### Example
+
+You may have a large group of servers that are partitioned into separate regions
+that correspond to actual geographic regions. Usually, you deploy to all of
+them, but there are cases where you want to deploy to a specific region.
+
+Capistrano recognizes the concept of a server's *role* and *hostname*, but has
+no concept of a *region*. In this case, you can construct your own filter that
+selects servers based on their region. When defining servers, you may provide
+them with a `region` property, and use that property in your filter.
+
+
+The filter could look like this:
+
+`config/deploy.rb`
+
+    class RegionFilter
+
+      def initialize(regions)
+        @regions = Array(regions)
+      end
+
+      def filter(servers)
+        servers.select {|server|
+          region = server.fetch(:region)
+          region && @regions.include?(region)
+        }
+      end
+
+    end
+
+You would add servers like this:
+
+`config/deploy/production.rb`
+
+    server('123.123.123.123', region: 'north-east')
+    server('12.12.12.12',     region: 'south-west')
+    server('4.5.6.7',         region: 'mid-west')
+
+To tell Capistrano to use this filter, you would use the
+`Configuration#add_filter` method. In this example, we look at the `REGIONS`
+environment variable, and take it to be a comma-separated list of regions that
+we're interested in:
+
+`config/deploy.rb`
+
+    if ENV['REGIONS']
+      regions = ENV['REGIONS'].split(',')
+      filter = RegionFilter.new(regions)
+      Capistrano::Configuration.env.add_filter(filter)
+    end
+
+We obtain a list of regions to deploy to from the environment variable,
+construct a new filter with those regions, and add it to Capistrano's list of
+filters.
+
+Of course, we're not limited to regions. Any time you can classify or partition
+a list of servers in a way that you only want to deploy to some of them, you can
+use a custom filter. For another example, you might arbitrarily assign your
+servers to either an *A* group or a *B* group, and deploy a new version only to
+the *B* group as a simple variant of A/B Testing.
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -110,8 +110,25 @@ module Capistrano
       @timestamp ||= Time.now.utc
     end
 
+    def add_filter(filter=nil, &block)
+      if block
+        raise ArgumentError, "Both a block and an object were given" if filter
+
+        filter = Object.new
+        def filter.filter(servers)
+          block.call(servers)
+        end
+      elsif !filter.respond_to? :filter
+        raise TypeError, "Provided custom filter <#{filter.inspect}> does " \
+                         "not have a public 'filter' method"
+      end
+      @custom_filters ||= []
+      @custom_filters << filter
+    end
+
     def setup_filters
-      @filters = cmdline_filters.clone
+      @filters = cmdline_filters
+      @filters += @custom_filters if @custom_filters
       @filters << Filter.new(:role, ENV["ROLES"]) if ENV["ROLES"]
       @filters << Filter.new(:host, ENV["HOSTS"]) if ENV["HOSTS"]
       fh = fetch_for(:filter, {}) || {}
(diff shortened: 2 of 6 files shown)


DylanFrese:
I would like to add tests for this, but the framework is very strange for someone who's only used JUnit.

docs/documentation/advanced-features/custom-filters/index.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+We obtain a list of regions to deploy to from the environment variable,
+construct a new filter with those regions, and add it to Capistrano's list of
+filters.
+
+Of course, we're not limited to regions. Any time you can classify or partition
+a list of servers in a way that you only want to deploy to some of them, you can
+use a custom filter. For another example, you might arbitrarily assign your
+servers to either an *A* group or a *B* group, and deploy a new version only to
+the *B* group as a simple variant of A/B Testing.
```
DylanFrese: I think the A/B example is pretty weak, a better real-world suggestion would be nice.

DylanFrese:
I've been using this repo https://github.com/DylanFrese/capistrano-custom-filter-test to test. A 'deploy' filters `release_roles(:all)` and prints out each server's hostname, then exits.

DylanFrese:
I can squash this if you want, @mattbrictson. I don't usually because it limits rerere/bisect and I think a finer history is always better.

lib/capistrano/configuration.rb
```diff
@@ -106,8 +106,17 @@ def timestamp
       @timestamp ||= Time.now.utc
     end
 
+    def add_filter(filter)
+      unless filter.respond_to? :filter
+        raise TypeError, "<#{filter}> is not a filter"
```
mattbrictson: I think a more descriptive error would be: `Custom filter does not have a public 'filter' method: #{filter}`.

docs/documentation/advanced-features/custom-filters/index.markdown
```diff
@@ -0,0 +1,77 @@
...
+
+Capistrano recognizes the concept of a server's *role* and *hostname*, but has
+no concept of a *region*. In this case, you can construct your own filter that
+selects servers based on their region. When defining servers, you may provide
+them with a `region` property, and use that property in your filter.
+
+
+The filter could look like this:
+
+`Capfile`
```
mattbrictson: Will your example still work if the code is placed in `deploy.rb` instead? Generally speaking, I would like the Capfile to be off-limits to everything except for `require` statements, with custom tasks and Ruby code placed in `deploy.rb` (or loaded from e.g. `lib`).

docs/documentation/advanced-features/custom-filters/index.markdown
mattbrictson: I think this works. This is already a very impressive set of documentation. 😄

docs/documentation/advanced-features/custom-filters/index.markdown
DylanFrese: Yep. So long as the custom filter is added before the first time filters are used. I put it in Capfile because that's (kind of) closer to where it would go in the projects I use Capistrano in.

lib/capistrano/configuration.rb
DylanFrese: Agreed.

DylanFrese:
As for tests, it would look like it would go in `spec/integration/dsl_spec.rb`, by proximity to similar tests.

mattbrictson:
@DylanFrese I think the proper way to test this is by adding two different tests.

First, I would suggest you add a test to `configuration_spec.rb` that calls `add_filter` with a dummy filter of your choosing, and then verifies that `filter(list)` uses that filter and produces the correct result. You should also have a test that passes at invalid object to `add_filter` and verifies that a `TypeError` is raised.

That would essentially be unit testing.

To be certain the feature works end-to-end, I also suggest adding a basic test to `integration/dsl_spec.rb`. It doesn't need to check the `TypeError` because that is already done by the unit test.

Does that make sense? I think you can probably use the existing tests in those files for a template for adding your own.

DylanFrese:
@mattbrictson Sorry it's taken so long to swing back around on this; I was sick last week.

I've added unit tests, as well as the ability to pass a block to `add_filter` in place of an object.

As for the integration test, I looked at `configuration_spec` for a while, and I'm not sure where exactly it fits, or if the test should go in more than one place. It would require mutating `Configuration` state, I believe.

Some of the integration tests there specify how `on` behaves when passed literal names, that is, if there is a role filter, `on` gets invoked on no servers (since none of the strings have a role associated with them). I am not sure of the correct behaviour here with regards to custom filters.

mattbrictson:
Sorry, haven't yet gotten a chance to review this PR further. If anyone is lurking and wants to jump in, please do! Otherwise I will circle back when I have more time.

docs/documentation/advanced-features/custom-filters/index.markdown
```diff
@@ -0,0 +1,83 @@
...
+
+      def initialize(regions)
+        @regions = Array(regions)
+      end
+
+      def filter(servers)
+        servers.select {|server|
+          region = server.fetch(:region)
+          region && @regions.include?(region)
+        }
```
will-in-wi: Generally, we'd use `do..end` syntax for multi-line blocks instead of curly braces.

mattbrictson:
The code and tests look good, but I'm having an issue where `rake spec` reports a handful of failures locally even though it passes on Travis. Can I get an extra pair of eyes on this, @will-in-wi or @leehambley?

@DylanFrese can you rebase on top of `master` and squash your commits? Thanks.

Here's what I'm seeing:

```
Failures:

  1) Capistrano::Svn::DefaultStrategy#check should test the repo url
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcd642f558>:0x007fdcd642f508 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/svn.rb:23:in `check'
     # ./spec/lib/capistrano/svn_spec.rb:41:in `block (3 levels) in <module:Capistrano>'

  2) Capistrano::Svn::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcd64159a0>:0x007fdcd6415950 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/svn.rb:17:in `test'
     # ./spec/lib/capistrano/svn_spec.rb:30:in `block (3 levels) in <module:Capistrano>'

  3) Capistrano::Git::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcdb2eae18>:0x007fdcdb2eadc8 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/git.rb:16:in `test'
     # ./spec/lib/capistrano/git_spec.rb:27:in `block (3 levels) in <module:Capistrano>'

  4) Capistrano::Hg::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcdb339d60>:0x007fdcdb339d10 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/hg.rb:14:in `test'
     # ./spec/lib/capistrano/hg_spec.rb:27:in `block (3 levels) in <module:Capistrano>'

  5) Capistrano::SCM#initialize should load the provided strategy
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcd4e4eef0>:0x007fdcd4e4ee50 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./spec/lib/capistrano/scm_spec.rb:21:in `test'
     # ./spec/lib/capistrano/scm_spec.rb:36:in `block (3 levels) in <module:Capistrano>'

  6) Capistrano::SCM Convenience methods #test! should return call test on the context
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fdcd4e1c5b8>:0x007fdcd4e1c518 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./spec/lib/capistrano/scm_spec.rb:46:in `block (4 levels) in <module:Capistrano>'
```

DylanFrese:
@mattbrictson Rebased.

mattbrictson:
After the rebase, all tests are now passing for me. I'll merge this in. Awesome!

@DylanFrese could you move your CHANGELOG entry to the top, under the `master` heading? It looks like it got put into the wrong place by the merge. After that I'll merge this PR. Thanks!

DylanFrese:
@mattbrictson Oops, fixed!

mattbrictson:
Thanks for the PR! ✨

PR #1781, opened 2016-09-27
create symlinks atomically by moving them in place

skaes:
On the legacy-v2 branch,symlinks are set up using two shell commands:
first rm; then ln -s. This creates a time window where the link doesn't exist.
During this time window passenger delivered a 403 response, our perl
apps delivered a 404. This patch fixes the problem.

### DIFF
--- a/lib/capistrano/recipes/deploy.rb
+++ b/lib/capistrano/recipes/deploy.rb
@@ -318,15 +318,17 @@ namespace :deploy do
     except `restart').
   DESC
   task :create_symlink, :except => { :no_release => true } do
+    tmp_current_path = File.join(releases_path, current_dir)
+
     on_rollback do
       if previous_release
-        run "#{try_sudo} rm -f #{current_path}; #{try_sudo} ln -s #{previous_release} #{current_path}; true"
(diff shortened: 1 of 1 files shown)


leehambley:
Thanks for the patch @skaes I believe we've been offered this patch more than one by a few people in the past but the `-n` was not portable if memory serves.

We'll look into that again. 

Is it possibly an option for you to run this as a patch in your local copy of Capistrano?

skaes:
Hi @leehambley, of course we can patch everything and you know I'm not shying away from patching something. However, not creating the link atomically is a real bug which can cause failed requests in production, so I thought to do the world a favor and submit a patch.

I can rewrite the patch to remove potentially existing old tmp_current links using a separate rm -f followed by a simple ln -s. What do you think?

skaes:
Another option is to let the deployment somehow and let the poor loser clean up.

Or just do rm -f beforehand.

Or just remove left over symlinks.

Or use a unique name for each release.

Just let me know.

will-in-wi:
Hey @skaes, sorry for sitting on this for so long. We're short-handed on the maintainer front. We should have taken care of this when it came in.

At this point, we have a policy of not supporting Capistrano 2 anymore. Capistrano 3 has been out for quite a while and is considered stable. I'm closing this PR for that reason.

PR #1782, opened 2016-10-03
Exclude docs from gem package to prevent gem from getting 10x larger

mattbrictson:
Now that the entire Capistrano docs website is contained in the docs directory, by default all those files will be included in the .gem file that is created by `rake build`.

This means that the next release of the capistrano gem will increase in size from 68 KB to 802 KB. More than 10x!

This commit updates the gemspec to include the docs directory when building the gem, to avoid this bloat.

Agree/disagree this is a good idea?

### DIFF
--- a/capistrano.gemspec
+++ b/capistrano.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |gem|
   gem.summary       = "Capistrano - Welcome to easy deployment with Ruby over SSH"
   gem.homepage      = "http://capistranorb.com/"
 
-  gem.files         = `git ls-files -z`.split("\x0")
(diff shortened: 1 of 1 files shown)


mattbrictson:
Oops, looks like `grep_v` is not available in older Rubies. Anyway, would still appreciate any strong opinions on whether this PR is worth pursuing.

leehambley:
I'd like to see it happen, being a lean Gem makes us a good citizen. 

Surely rather than `#grep_v` we can use something like:

```
.grep_v(/^docs/)
.reject { |f| /^docs/ ~= f }
```

?

mattbrictson:
👆 Good idea. Fixed.

PR #1787, opened 2016-10-13
Allow the tar command name to be overridden

caius:
When using the git integration, it generates a tarball and then untars it to generate the release folder. This has been hardcoded to use `tar` up until now, but some OS's provide a tar that isn't gnu compatible (eg, SmartOS). Instead they generally have gnu-compatible tar available under a different name (e.g., `gtar`).

This is now picked up from a variable, defaulting to `tar` for backwards compatibility and can be overridden when required:

    set :tar, 'gtar'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.6.1...HEAD
 
 * Your contribution here!
+* The `tar` used by the Git SCM now honors the SSHKit command map, allowing an alternative tar binary to be used (e.g. gtar) #1787 (@caius)
 * Fix test suite to work with Mocha 1.2.0 (@caius)
 * `remote_file` feature has been removed and is no longer available to use @SaiVardhan
 * Fix bug where host_filter and role_filter were overly greedy [#1766](https://github.com/capistrano/capistrano/issues/1766) (@cseeger-epages)
--- a/lib/capistrano/git.rb
+++ b/lib/capistrano/git.rb
@@ -41,9 +41,9 @@ class Capistrano::Git < Capistrano::SCM
       if (tree = fetch(:repo_tree))
         tree = tree.slice %r#^/?(.*?)/?$#, 1
         components = tree.split("/").size
-        git :archive, fetch(:branch), tree, "| tar -x --strip-components #{components} -f - -C", release_path
+        git :archive, fetch(:branch), tree, "| #{SSHKit.config.command_map[:tar]} -x --strip-components #{components} -f - -C", release_path
       else
-        git :archive, fetch(:branch), "| tar -x -f - -C", release_path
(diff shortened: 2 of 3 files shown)


mikegee:
Cap 2 used to allow overriding `copy_remote_tar` and `copy_local_tar`:

https://github.com/capistrano/capistrano/blob/v2.15.9/lib/capistrano/recipes/deploy/strategy/copy.rb#L297-L298

I'm not completely sure where `tar` is running in this PR. Consider naming it `remote_tar`, for instance, if it is executed remotely.

mattbrictson:
Should we leverage the command map for this instead? /cc @leehambley @will-in-wi

caius:
Can we use the command map for piped commands? I wasn't sure if that would work.

Happy to rework this if that's possible 🙂

mattbrictson:
I'm not sure it would work in all cases, but could you do this:

```ruby
SSHKit.config.command_map[:tar]
```

instead of:

```ruby
fetch(:tar, 'tar')
```

?

mattbrictson:
This PR is blocked by #1789, which is causing Travis to fail. Can someone take a look?

mattbrictson:
A safer approach might be to follow this example: https://github.com/lavmeiker/capistrano-wpcli/pull/24/files

Basically, split the `git archive | tar ...` command into separate commands with a tempfile, such that `:tar` can be its own command (and thus substitutable using the command map).

Thoughts?

Also, be aware that Travis is failing for this PR due to RuboCop violations. Be sure to run `bundle exec rake` to verify tests pass and RuboCop validates.

leehambley:
> Basically, split the git archive | tar ... command into separate commands with a tempfile, such that :tar can be its own command (and thus substitutable using the command map).

Using a tempfile will increase the time taken to export significantly for large repositories, that's why it's a pipe presently, it was a problem for the customer for whom I was working whilst I build this feature.

The command map is the right way to solve this, and the hard-coded strings could easily lookup the command e.g

``` ruby
execute "#{SSHKit.config.command_map[:git]} ..... | #{SSHKit.config.command_map[:tar]} ....."
```

caius:
Ah cool, thanks for that @leehambley. I'll get it fixed up accordingly

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.6.1...HEAD
 
 * Your contribution here!
+* Allow setting `tar` binary name (defaults to `tar` as before) (@caius)
```
mattbrictson: Can you make this a bit more descriptive? How about:

> The `tar` used by the Git SCM now honors the SSHKit command map, allowing an alternative tar binary to be used (e.g. `gtar`)  #1787

caius:
Sure thing, forgot about that. Updated 😄

mattbrictson:
Nice! Thanks for the PR 🙇

PR #1790, opened 2016-10-13
Fix test suite for mocha 1.2.0

### LINKED ISSUE #1789: Latest version of mocha breaks Capistrano specs
Mocha 1.2.0 was just released. Updating causes Capistrano specs to fail.

### Steps to reproduce

1. In Capistrano master
2. `bundle update`
3. `bundle exec rake`

### Expected behavior

All specs should pass.

### Actual behavior

```
Failures:

  1) Capistrano::Hg::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d241a2b8>:0x007fb6d241a218 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/hg.rb:14:in `test'
     # ./spec/lib/capistrano/hg_spec.rb:27:in `block (3 levels) in <module:Capistrano>'

  2) Capistrano::SCM#initialize should load the provided strategy
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d3051c48>:0x007fb6d3051bf8 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./spec/lib/capistrano/scm_spec.rb:21:in `test'
     # ./spec/lib/capistrano/scm_spec.rb:36:in `block (3 levels) in <module:Capistrano>'

  3) Capistrano::SCM Convenience methods #test! should return call test on the context
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d2533dc0>:0x007fb6d2533d70 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./spec/lib/capistrano/scm_spec.rb:46:in `block (4 levels) in <module:Capistrano>'

  4) Capistrano::Svn::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d2613e20>:0x007fb6d2613dd0 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/svn.rb:17:in `test'
     # ./spec/lib/capistrano/svn_spec.rb:30:in `block (3 levels) in <module:Capistrano>'

  5) Capistrano::Svn::DefaultStrategy#check should test the repo url
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d25e3540>:0x007fb6d25e34f0 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/svn.rb:23:in `check'
     # ./spec/lib/capistrano/svn_spec.rb:41:in `block (3 levels) in <module:Capistrano>'

  6) Capistrano::Git::DefaultStrategy#test should call test for repo HEAD
     Failure/Error: context.test(*args)

     NoMethodError:
       private method `test' called for #<#<Class:0x007fb6d25687f0>:0x007fb6d2568750 @mocha=nil>
     # ./lib/capistrano/scm.rb:27:in `test!'
     # ./lib/capistrano/git.rb:16:in `test'
     # ./spec/lib/capistrano/git_spec.rb:27:in `block (3 levels) in <module:Capistrano>'

Finished in 0.84893 seconds (files took 1.04 seconds to load)
361 examples, 6 failures, 2 pending
```

caius:
This comes down to a behaviour change in mocha not playing so nicely with how we've been generating empty mock objects so far.

Mocha 1.2.0 now pays attention to the visibility of methods you stub (expect(), etc) and disallows you calling them in a different scope. ie, if you stub a private method it can't be called from outside the instance. Previously doing that worked fine and you had no indication you were calling a private method. This change totally makes sense!

However, we're providing "empty" contexts in our specs by using `Class.new.new` which is creating an anonymous instance of an anonymous class. Note that the anonymous class contains Kernel in the ancestors.

We then stub out `test` on these anonymous instances, which we don't expect to respond to `test` without the stubbing, and the stubbed method then gets called. However, remember back to where we had `Kernel` methods available in our anonymous instance? Turns out there's `Kernel#test` and it's marked private! Except that our method on the real context instances is public, so we want that to work.

Easiest fix I can see is to use Mocha's `mock` method to create a mock object for us to stub out our methods on, rather than trying to be ruby-ish and use an anonymous instance of an anonymous class. (I attempted using BasicObject first, but we'd have to include Mocha into that ourselves, and it leans on some methods provided by Kernel.)

Fixes #1789

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.6.1...HEAD
 
 * Your contribution here!
+* Fix test suite to work with Mocha 1.2.0 (@caius)
 * `remote_file` feature has been removed and is no longer available to use @SaiVardhan
 * Fix bug where host_filter and role_filter were overly greedy [#1766](https://github.com/capistrano/capistrano/issues/1766) (@cseeger-epages)
 
--- a/spec/lib/capistrano/git_spec.rb
+++ b/spec/lib/capistrano/git_spec.rb
@@ -4,7 +4,7 @@ require "capistrano/git"
 
 module Capistrano
   describe Git do
-    let(:context) { Class.new.new }
+    let(:context) { mock }
     subject { Capistrano::Git.new(context, Capistrano::Git::DefaultStrategy) }
 
     describe "#git" do
@@ -16,7 +16,7 @@ module Capistrano
   end
 
   describe Git::DefaultStrategy do
-    let(:context) { Class.new.new }
+    let(:context) { mock }
     subject { Capistrano::Git.new(context, Capistrano::Git::DefaultStrategy) }
 
     describe "#test" do
--- a/spec/lib/capistrano/hg_spec.rb
+++ b/spec/lib/capistrano/hg_spec.rb
@@ -4,7 +4,7 @@ require "capistrano/hg"
 
 module Capistrano
   describe Hg do
-    let(:context) { Class.new.new }
+    let(:context) { mock }
     subject { Capistrano::Hg.new(context, Capistrano::Hg::DefaultStrategy) }
 
     describe "#hg" do
@@ -16,7 +16,7 @@ module Capistrano
   end
 
   describe Hg::DefaultStrategy do
-    let(:context) { Class.new.new }
+    let(:context) { mock }
     subject { Capistrano::Hg.new(context, Capistrano::Hg::DefaultStrategy) }
 
     describe "#test" do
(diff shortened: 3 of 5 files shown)


mattbrictson:
Excellent detective work! Thanks ✨

PR #1793, opened 2016-10-18
Rename git to git_plugin, etc.

mattbrictson:
### Summary

This makes it more explicit which object we are referring to, which is to say the SCM plugin object. As suggested by @will-in-wi here: https://github.com/capistrano/capistrano/pull/1572/files#r54176503

### Short checklist

- [ ] ~~If relevant, did you create a test?~~
- [ ] ~~If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry?~~


will-in-wi:
LGTM 👍

PR #1795, opened 2016-10-24
Set up Danger!

mattbrictson:
- [ ] ~~If relevant, did you create a test?~~
- [ ] ~~If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry?~~

"Danger" is a gem that automatically runs certain sanity checks against GitHub pull requests. Whenever a Capistrano PR is opened, Danger will check whether a CHANGELOG entry was included, and whether tests were included, and add a comment to the PR if there are violations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: ruby
 rvm:
-  - 2.3.0
+  - 2.3.1
   - 2.2
   - 2.1
   - 2.0
@@ -13,6 +13,10 @@ matrix:
   include:
     - rvm: rbx-2
       script: bundle exec rake spec
+    # Run Danger only once, on 2.3.1
+    - rvm: 2.3.1
+      before_script: bundle exec danger
+
 script: bundle exec rake spec rubocop
 install: bundle install --jobs=1
 cache: bundler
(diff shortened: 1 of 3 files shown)


leehambley:
What key does Danger use from Travis to post to GitHub? I presume it uses the same key Travis uses to clone?

mattbrictson:
@leehambley I added the key for the @capistrano-bot to the Travis ENV. You can see it in the UI here: https://travis-ci.org/capistrano/capistrano/settings . This allows @capistrano-bot to post comments on PRs. Travis itself doesn't use a key for cloning since the repo is public.

mattbrictson:
I have Danger already running on the Airbrussh repo, and it seems to be working. I think it will save us some time by enforcing CHANGELOG entries in particular, which otherwise lengthen the PR review process. Let me know what you think!

mattbrictson:
@leehambley Would be nice to have Danger pester people about adding CHANGELOG entries, rather than have to do it ourselves for every PR. Can we enable this?

leehambley:
Sorry, I thought we'd merged this - let's do it!

Issue #1792, opened 2016-10-18
Make SCM plugin registration more explicit


mattbrictson:
Per @will-in-wi's [comments](https://github.com/capistrano/capistrano/pull/1572/files#r54174714) originally in #1572:

```ruby
# This is somewhat of a hack, because there is no guarantee that a third-
# party SCM will necessarily implement the typical SCM tasks. But it works
# well enough for the built-in SCMs.
def scm_plugin_loaded?
  Rake::Task.tasks.any? { |t| t.name =~ /^[^:]+:create_release/ }
```

> Would it make more sense to require that scm plugins register themselves to avoid this? This feels like it could lead to some difficult to debug situations for a developer creating a new scm plugin. Or possibly throw a warning if fetch(:scm) is set and there are no such tasks.

@mattbrictson:

> Yeah, this is tricky. I guess registration is unavoidable. How should SCM plugins register themselves? Do you think all SCM plugins should inherit from a Plugin::SCM parent class, perhaps? Or add metadata to the plugin that describes what it does, like provides :scm?

@will-in-wi 

> Can we take advantage of the Capistrano::Git.new stanza to have it register itself? IIRC, elsewhere in this change, it tracks things that were called that way…

mattbrictson:
One option is to add a new registration method, `install_scm`. In other words, an SCM would be activated like this:

```ruby
require "capistrano/scm/git"
install_scm Capistrano::SCM::Git
```

It works the same as `install_plugin`, except it sets an extra bit of state that says "an SCM has been registered". The SCMResolver can know for sure that the user has selected an SCM instead of using an error-prone heuristic.

Alternatively, we introduce a `Capistrano::SCM::Plugin` base class that all SCMs must inherit from. That would make it the plugin developer's responsibility to write the plugin correctly, rather than the end-user's responsibility to register it correctly. This solution might be less finicky.

Possible implementation:

```ruby
class Capistrano::Plugin < Rake::TaskLib
  # ...
  def scm?
    @scm
  end
end

class Capistrano::SCM::Plugin < Capistrano::Plugin
  def initialize
    @scm = true
  end
end
```

Thus by inheriting from `Capistrano::SCM::Plugin` you automatically get the `scm? => true` behavior, which is the hint that SCMResovler needs. But at the same time, you don't *have* to inherit as long as you respond to the `scm?` message; any object can be an SCM plugin.

will-in-wi:
[...]

PR #1797, opened 2016-10-25
Change SCMs to inherit from new SCM::Plugin base class

mattbrictson:
### Summary

As proposed in #1792, this makes SCM plugin registration more explicit. Now, all SCMs inherit from `SCM::Plugin`, which allows Capistrano to know for sure whether an SCM is loaded. This means we can remove a hack that `SCMResolver` was previously using.

### Short checklist

This is a work in progress. @will-in-wi  does it look good so far?

### DIFF
--- a/spec/lib/capistrano/configuration/plugin_installer_spec.rb
+++ b/spec/lib/capistrano/configuration/plugin_installer_spec.rb
@@ -1,0 +1,98 @@
+require "spec_helper"
+require "capistrano/plugin"
+require "capistrano/scm/plugin"
+
+module Capistrano
+  class Configuration
+    class ExamplePlugin < Capistrano::Plugin
+      def set_defaults
+        set_if_empty :example_variable, "foo"
+      end
+
+      def define_tasks
+        task :example
+        task :example_prerequisite
+      end
+
+      def register_hooks
+        before :example, :example_prerequisite
+      end
+    end
+
+    class ExampleSCMPlugin < Capistrano::SCM::Plugin
+    end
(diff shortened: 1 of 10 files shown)


will-in-wi:
I like this a lot. 👍

mattbrictson:
Great! I'll work on tests and a CHANGELOG so this can be merged. Then I think we can ship 3.7.0!

leehambley:
Really excited about this @mattbrictson, nice work, it's only been 10 years in the coming :)

mattbrictson:
I added tests so I think this is ready to go. LGTM?

mattbrictson:
(Also, I omitted a CHANGELOG entry since this is still part of the bigger SCM refactor that is already documented for 3.7.0)

spec/lib/capistrano/configuration/plugin_installer_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      end
+
+      def define_tasks
+        task :example do
+          # intentionally blank
+        end
+
+        task :example_prerequisite do
+          # intentionally blank
+        end
```
will-in-wi: You could shorten these with `task(:example) {}`, but I'm not sure if that's any better.

mattbrictson: I shortened it to

```ruby
task :example
```

(apparently the block is optional)

mattbrictson:
Merging! 🎉

PR #1804, opened 2016-11-17
Suppress log by filtering remote refs

aeroastro:
### Summary

This patch can suppress long logs generated by `git ls-remote`, especially for repositories with many tags and branches.

Because the primary objective of `git ls-remote` used is to check the access rights to `remote_url`, we can omit a number of remote refs by specifying HEAD as `<refs>` at the end of the command.

Although the default exit code of `git ls-remote` is always 0 regardless of matched refs, I think specifying `HEAD` as `<refs>` is fairly simple option for Git users.

### Short checklist

### Other Information

Ref. https://git-scm.com/docs/git-ls-remote.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.7.0.beta1...HEAD
 
 * Your contribution here!
+* Suppress log messages of `git ls-remote` by filtering remote refs (@aeroastro)
 
 ## `3.7.0.beta1` (2016-11-02)
 
--- a/features/deploy.feature
+++ b/features/deploy.feature
@@ -7,7 +7,6 @@ Feature: Deploy
   Scenario: Creating the repo
     When I run cap "git:check"
     Then the task is successful
-    And references in the remote repo are listed
     And git wrapper permissions are 0700
 
   Scenario: Creating the directory structure
--- a/lib/capistrano/scm/git.rb
+++ b/lib/capistrano/scm/git.rb
@@ -32,7 +32,7 @@ class Capistrano::SCM::Git < Capistrano::SCM::Plugin
   end
 
   def check_repo_is_reachable
-    git :'ls-remote --heads', repo_url
+    git :'ls-remote', repo_url, "HEAD"
   end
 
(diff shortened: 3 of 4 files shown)


leehambley:
Thanks for the contribution @aeroastro - I have a question:

- which versions of Git support the change you made?

First impression is that it looks OK, since it's only a repo-check and doesn't need to enumerate all objects, we could definitely pull fewer tags/heads/etc, and still have equal functionality.

aeroastro:
Thank you for the comment.

> which versions of Git support the change you made?

This change support Git with version later than or equal to 0.99.4.

The `<repository> <refs>...` usage was first implemented by https://github.com/git/git/commit/972b6fe746f0d9a3ea13118d80d9371a9ff68dba and officially released by 0.99.4 https://github.com/git/git/commit/bf570303153902ec3d85570ed24515bcf8948848 on 11 August 2005. On the other hand, `git ls-remote` command itself was released by 0.99.2.

leehambley:
Thanks, great news. If you;d like to be credited in the `CHANGELOG`, please just adjust the PR, otherwise let me know, and I'll just merge it.

mattbrictson:
@aeroastro Can you add a CHANGELOG entry? Then we can merge this PR. Thanks!

aeroastro:
Sorry for the late response and thank you for the review.
I added a CHANGELOG entry. 🐱

mattbrictson:
`features/deploy.feature` is failing in this PR. I think at least one of the assertions is now invalid; specifically, "references in the remote repo are listed" is no longer true. Can you remove that assertion?

Unfortunately the only way to check Cucumber features is to run them locally, since we can't do it in Travis. Assuming you have a working Vagrant setup, you can run the Cucumber features using:

```
gem install bundler -v 1.10.5
bundle _1.10.5_ exec rake features
```

aeroastro:
I removed the assertion on "And references in the remote repo are listed".

Then, by executing `bundle _1.10.5_ exec rake features`, it was confirmed that all the scenarios and steps passed except for 1 pending scenario and 1 pending step.

mattbrictson:
Excellent work. Thanks for the PR! 🙇

aeroastro:
Thank you! 🐱

Issue #1545, opened 2015-12-29
Detect changed repo_url


marcovtwout:
Using svn, when I change the repo_url, the clone on the server is not changed and the old version remains.

The solution is simple: when the repo_url changes, I expect the repo clone to be deleted and checked out again.

marcovtwout:
I found a plugin that solves this, but I think this should be core behavior: https://gist.github.com/bluefuton/982691

leehambley:
If you prepare a PR with tests for incorporating this behaviour into Capistrano, we'd be glad. Actually Git suffers the same problem sometimes, and it can be very confusing for people.

marcovtwout:
Implementing that properly is a bit out of my reach I'm afraid (I'm new to Ruby development).

mattbrictson:
Even if we're not building the feature right now, it could be useful to discuss it a little more, for future reference.

The [Gist you mentioned](https://gist.github.com/bluefuton/982691) seems to delete the entire repository checkout if it finds the `repo_url` has changed, so that the repo can be re-cloned. I guess that works, but I feel nervous about Capistrano doing a wholesale delete with `rm -rf` automatically without user confirmation.

Maybe a safer solution is for Capistrano to abort with an error message like this:

```
It seems that your repo_url has changed since your last deployment.
Run `cap production [scm]:reset` to delete Capistrano's cache and fetch the new repo_url.
```

And then an e.g. `svn:reset`, `git:reset`, etc. would delete the checkout and do a fresh clone.

leehambley:
I'm about to support a customer (this week) who has a similar problem for the first time, I'll see if I can come up with the bones of a sane task to use as a reference for discussing this further

mattbrictson:
Possibly consider addressing this as part of the same solution: #1482 (in this case repo_url hasn't changed, but cached credentials have gone stale).

leehambley:
> I'm about to support a customer (this week) who has a similar problem for the first time, I'll see if I can come up with the bones of a sane task to use as a reference for discussing this further

Ran into some problems, so we nuked the repo - since we "own" `repo` in the `:deploy_to` it seems relatively harmless, but I do think we should do as you suggested, and prompt.

mattbrictson:
Although a problem with prompting is if you are running Capistrano in non-interactive settings, like CI…

marcovtwout:
> since we "own" repo in the :deploy_to it seems relatively harmless

[...]

PR #1807, opened 2016-11-26
Add a warning about changing :repo_url

Mingan:
This PR adds a warning to docs about how to proceed after changing `:repo_url` until #1545 gets solved.

### DIFF
--- a/docs/documentation/getting-started/configuration/index.markdown
+++ b/docs/documentation/getting-started/configuration/index.markdown
@@ -68,6 +68,7 @@ The following variables are settable:
   * Example: `set :repo_url, 'git@example.com:me/my_repo.git'` for a git repo located in /home/git/me
   * Hint #1: to access a repo on a machine using a non-standard ssh port: `set :repo_url, 'ssh://git@example.com:30000/~/me/my_repo.git'`
   * Hint #2: when using :svn and branches, declare the repo_url like this: `set :repo_url, -> { "svn://myhost/myrepo/#{fetch(:branch)}" }`
+  * Warning: if you move your repository to a new URL and change this variable, already deployed remote servers won't reflect this change automatically, you have to manually re-configure the repository on the remote servers (in path determined by `:repo_path`) or delete it and let Capistrano recreate it on the next deploy using the updated URL.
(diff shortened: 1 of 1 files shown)


docs/documentation/getting-started/configuration/index.markdown
```diff
@@ -68,6 +68,7 @@ The following variables are settable:
   * Example: `set :repo_url, 'git@example.com:me/my_repo.git'` for a git repo located in /home/git/me
   * Hint #1: to access a repo on a machine using a non-standard ssh port: `set :repo_url, 'ssh://git@example.com:30000/~/me/my_repo.git'`
   * Hint #2: when using :svn and branches, declare the repo_url like this: `set :repo_url, -> { "svn://myhost/myrepo/#{fetch(:branch)}" }`
+  * Warning: if you move your repository to a new URL and change this variable, already deployed remote servers won't reflect this change automatically, you have to manually re-configure the repository on the remote servers (in path determined by `:repo_path`) or delete it and let Capistrano recreate it on the next deploy using the updated URL.
```
will-in-wi: Might be worth saying how to delete it (aka delete the repo folder).

Mingan: That's the idea I had when I went into it. Then I thought about it and realized that you could only do with changing the repo (`git origin set-url`) which might be worth considering if your repo is huge and you want to avoid re-downloading all of it. On top of that I noticed that `:repo_path` is a configurable variable so saying `rm -rf repo` might not apply.

As a compromise, I'd go with `... or delete it (rm -rf repo in the default setup) and let Capistrano...`

will-in-wi: The compromise option seems like a good one to me. It'll help the normal case (which is the majority of people, I think…) and if someone has done something advanced, hopefully they'll know what they are doing with this.

will-in-wi:
LGTM! @leehambley Merge?

mattbrictson:
Looks good. Thanks for the docs! 🙇

Mingan:
My pleasure

Issue #1810, opened 2016-12-02
Removing old releases goes wrong


markri:
When removing old releases (default behaviour, and explicitly setted at :keep_releases: 5) it removes the newly created release instead of the old releases.

Looking at lib/capistrano/tasks/deploy.rake @ line 151 it says

`releases = capture(:ls, "-xtr", releases_path).split`

Which should be an ordered list, but executing it on command line shell it give me
```
ls -xtr
20161201100437  20161201112702  20161202132042  20161202131122  20161201114136
```

For more easy viewing I added some line breaks

```
20161201100437  
20161201112702  
20161202132042  
20161202131122  
20161201114136
```
Which is definitely not in order, which is why it removes the wrong releases at cleanup. 
Also executed:

```
[marco@server releases]$ ls -alsrt
total 24
4 drwxr-xr-x. 13 deploy deploy 4096  1 dec 11:00 20161201100437
4 drwxr-xr-x. 13 deploy deploy 4096  1 dec 12:25 20161201112702
4 drwxr-xr-x. 13 deploy deploy 4096  1 dec 12:39 20161202132042
4 drwxr-xr-x. 13 deploy deploy 4096  1 dec 12:39 20161202131122
4 drwxr-xr-x. 13 deploy deploy 4096  1 dec 12:39 20161201114136
0 drwxrwxr-x.  4 deploy deploy   68  2 dec 14:20 ..
4 drwxrwxr-x.  7 deploy deploy 4096  2 dec 14:20 .
[marco@server releases]$ date
vr dec  2 14:50:24 CET 2016
```

As it seems the release it self has a different creation date (yesterday) than the actual constructed timestamp (e.g. file system date is 1 dec 12:39 while folder timestamp is 2016 12 02 13:11:22), how is this possible?

Used gems:
```
capistrano (3.6.1)
capistrano-composer (0.0.6)
capistrano-file-permissions (0.1.1)
capistrano-harrow (0.5.3)
capistrano-symfony (0.4.0)
capistrano-tarball (0.1.2)
```

Capfile
```
require "capistrano/setup"
require "capistrano/deploy"
require 'capistrano/symfony'
require 'capistrano/tarball/load'

```

deploy.rb
```
lock '3.6.1'
set :application, ENV['project']
set :deploy_to, ENV['caproot']
set :scm, :tarball
set :tarball_release_tag, ENV['revision']
set :tarball_name, "tarbal.tar.gz"
[...]

PR #1811, opened 2016-12-05
Fixes Bug #1810 removal of old releases

markri:


leehambley:
LGTM, I find the guard against unknown dir names to be robust enough for a feature I'm not deeply invested in.

Can merge when ready, if I get a +1 from Matt or Will

will-in-wi:
This looks good to me! We probably need a changelog entry.

lib/capistrano/tasks/deploy.rake
```diff
@@ -148,8 +148,10 @@ namespace :deploy do
   desc "Clean up old releases"
   task :cleanup do
     on release_roles :all do |host|
-      releases = capture(:ls, "-xtr", releases_path).split
-      if releases.count >= fetch(:keep_releases)
+      releases = capture(:ls, "-x", releases_path).split
```
mattbrictson: Since you have removed the `-r` flag, doesn't that mean the releases are no longer in reverse order? If so, the Ruby code for choosing which releases to delete needs to be flipped, right?

mattbrictson: Ah, you're right. Make sense.

mattbrictson:
Please add a CHANGELOG entry, and then this will be good to merge. Thanks! ✨

mattbrictson:
Looks good. Thanks again! 🙇

markri:
np, glad I could help :-)

PR #1825, opened 2016-12-27
Document the additional gems needed to support ed25519 keys

mattbrictson:
Through trial and error, I discovered I needed to upgrade to net-ssh 4.0.0, plus install three additional optional gems in order to use the new, more secure ed25519 type SSH keys with Capistrano.

Is the Capistrano README the best place to document this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,12 @@ Add Capistrano to your project's Gemfile:
 ``` ruby
 group :development do
   gem "capistrano", "~> 3.7"
+
+  # Optional: these gems are needed to support ed25519 SSH keys
+  gem "bcrypt_pbkdf", require: false
+  gem "net-ssh", ">= 4.0.0", require: false
+  gem "rbnacl", "~> 3.4", require: false
+  gem "rbnacl-libsodium", require: false
 end
 ```
 


will-in-wi:
This seems reasonable to me.

will-in-wi:
We may want to consider just requiring this whole cloth into the next version of Capistrano. CentOS 7 uses Elliptic Curve keys by default now.

leehambley:
I'd be a fan of adding something akin to:

> Capistrano is a high level wrapper which relies on many Gems for support for protocols and specific authentication schemes. It is *strongly* recommended to examine the documentation for SSHKit and Net:SSH as you become familiar with Capistrano. For example, only through Net:SSH's own documentation can you learn which Gems are required for the newest category of elliptical curve SSH key authentication.

My proposed text sounds a bit "Fu. RTFM" but we have always had to strike a balance between informing people directly and directing them to docs for their shell, operating system, build tool of choice, etc.

Alternative is we just depend on their suggested gems explicitly (in this case, though I'd think we should add that dependency to SSHKit, actually). Sometimes a mixed blessing that Rubygems doesn't have *suggested* packages like Aptitude, etc.

mattbrictson:
> Alternative is we just depend on their suggested gems explicitly (in this case, though I'd think we should add that dependency to SSHKit, actually)

I think the best long-term solution is for net-ssh to make them hard dependencies. I am not sure why they are an optional dependency right now. Maybe it is out of caution until the new ed25519 support is proved out and becomes more widespread.

leehambley:
> Maybe it is out of caution until the new ed25519 support is proved out and becomes more widespread.

I think two reasons, one in a comment when they're discussing the RC release they mention 

> For now rbnacl-libsodium, rbnacl, bcrypt_pbkdf have to be added to Gemfile for ed25519 to work.

and they're allowing an env var to skip adding them as dev dependencies:

> https://github.com/net-ssh/net-ssh/blob/master/net-ssh.gemspec#L31-L35

I think one problem here is that it adds a dependency on a C library (`libsodium`) via the `rbnacl` gem which complicates everything for what was, until now a pure Ruby solution.

I've temporarily, and perhaps impolitely hijacked their issue https://github.com/net-ssh/net-ssh/issues/214 to ask what their long-term plans are for requiring those Gems directly

mattbrictson:
With https://github.com/net-ssh/net-ssh/pull/479 merged, net-ssh now prints exactly what dependencies it requires. Perhaps that is good enough for us and we don't need to update the Capistrano docs with this PR?

```
$ BUNDLE_GEMFILE=./Gemfile.norbnacl bundle exec irb
2.3.0 :001 > require 'net/ssh'
 => true 
2.3.0 :002 > Net::SSH.start('localhost','foo', {keys: 'test/keys/id_ed25519_pwd'})
NotImplementedError: unsupported key type `ssh-ed25519'
net-ssh requires the following gems for ed25519 support:
 * rbnacl (>= 3.2, < 5.0)
 * rbnacl-libsodium, if your system doesn't have libsodium installed.
 * bcrypt_pbkdf (>= 1.0, < 2.0)
See https://github.com/net-ssh/net-ssh/issues/478 for more information
Gem::LoadError : "rbnacl is not part of the bundle. Add it to Gemfile."
```

PR #1826, opened 2016-12-28
Allow repo_url changes to be realized on existing hosts when using the Git SCM

javanthropus:
When the repo_url setting needs to be changed, currently exiting deployment hosts will not update their mirror URL to fetch from the new URL.  The user either needs to clobber the mirror directory or manually update the remote URL prior to deployment.

This patch sets the URL just before updating from the origin.  It's effectively a fast and harmless noop most of the time, but we have a large number of hosts for various applications that have needed this operation as we have moved from one Git host to another.

### DIFF
--- a/lib/capistrano/scm/git.rb
+++ b/lib/capistrano/scm/git.rb
@@ -44,6 +44,9 @@ class Capistrano::SCM::Git < Capistrano::SCM::Plugin
   end
 
   def update_mirror
+    # Update the origin URL if necessary.
+    git :remote, "set-url", "origin", repo_url
+
     # Note: Requires git version 1.9 or greater
     if (depth = fetch(:git_shallow_clone))
       git :fetch, "--depth", depth, "origin", fetch(:branch)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ https://github.com/capistrano/capistrano/compare/v3.7.1...HEAD
 
 * Your contribution here!
 * Suppress log messages of `git ls-remote` by filtering remote refs (@aeroastro)
+* The Git SCM now allows the repo_url to be changed without manually wiping out the mirror on each target host first (@javanthropus)
 
 ## `3.7.1` (2016-12-16)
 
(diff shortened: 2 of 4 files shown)


mattbrictson:
Thanks for the PR! I like that this seems to be a simple, elegant solution to part of the `repo_url` problem described in #1545. It won't work for svn, and it probably won't work if you are using git over HTTP and the credentials have changed, but it is a step in the right direction.

As part of this PR, could you please also update the docs that were changed in #1807?

javanthropus:
I'll update the doc to mention that SCMs other than Git continue to have the relocation problem.

It's also possible to relocate an SVN workspace: http://stackoverflow.com/a/16948970.  The relocate subcommand could be applied similarly to my change for Git.

javanthropus:
Is there anything else I need to do to get this accepted?

mattbrictson:
@javanthropus Yes, could you rebase this on top of master and resolve the conflict?

@leehambley or @will-in-wi could I get a second pair of eyes on this PR? Thanks!

lib/capistrano/scm/git.rb
```diff
@@ -44,6 +44,9 @@ def clone_repo
   end
 
   def update_mirror
+    # Update the origin URL if necessary.
+    git :remote, "set-url", "origin", repo_url
```
leehambley: I looked to see if `git-remote set-url` had a long or short history, to see if we needed to address `git` version requirements at all. Spoiler Alert: it seems to have been around basically for **at least** [three years, since `1.8.2.3`][forever], so all fine.

This looks like a simple enough change, I'd be glad to have it in. 

[forever]: https://git-scm.com/docs/git-remote/1.8.2.3

javanthropus:
I rebased to remove the merge conflict with master.  The rbx test failure doesn't seem to be related to my changes at all.

mattbrictson:
Looks good, thanks! I don't know why rbx is suddenly causing a build failure; we explicitly tell Travis to ignore those errors. I agree it is unrelated to the changes in the PR.

Issue #1824, opened 2016-12-24
Cap fails to start when SCM plugin is installed too early in Capfile


mattbrictson:
Originally reported on Stack Overflow here: http://stackoverflow.com/q/41307044/4625365

### Steps to reproduce

1. Generate a Capfile and supporting files using `cap install`.
2. Edit the resulting Capfile so that `install_plugin Capistrano::SCM::Git` comes before `require "capistrano/deploy"` (see example below).
3. Try to execute a `cap` command.

```
$ cap production doctor
(Backtrace restricted to imported tasks)
cap aborted!
Don't know how to build task 'deploy:new_release_path' (see --tasks)
```

Capfile:

```ruby
require "capistrano/setup"
require "capistrano/scm/git"
install_plugin Capistrano::SCM::Git
require "capistrano/deploy"
Dir.glob("lib/capistrano/tasks/*.rake").each { |r| import r }
```

### Expected behavior

Ideally the Capfile would be forgiving of `require` ordering. If this is not practical, then the `cap` command should at least emit a helpful error message that allows the user to easily identify and fix the problem.

### Analysis

The SCM plugins use `after "deploy:new_release_path", ...` to install their behavior. If the deploy tasks haven't been loaded yet, this fails.

We could perhaps:

* Allow `after` to register hooks even for non-existent tasks (i.e. defer the hook until the task is defined later). Downside: if you make a typo in a task name, you no longer get an error.
* Add `require "capistrano/deploy"` to the top of `git.rb`, since it is technically a dependency. Downside: someone someday may want to use the Git plugin but provide their own deploy tasks instead of the default Capistrano ones; this change would couple them together.
* Make no changes at all; this is user error.
* Catch the exception and somehow print a better error message for this scenario.

mattbrictson:
An additional aspect to this issue: When upgrading an existing project to 3.7.0, the user is given this instruction:

```
[Deprecation Notice] `set :scm, :git` is deprecated.
To ensure your project is compatible with future versions of Capistrano,
remove the :scm setting and instead add these lines to your Capfile:

    require "capistrano/scm/git"
    install_plugin Capistrano::SCM::Git
```

The message does not say *where* in the Capfile these lines should be added, so it is understandable that users could make a mistake.

mcblum:
[...]

PR #1828, opened 2017-01-03
Capistrano::SCM::* depends on Capistrano/deploy

glen0071:
Editing the UPGRADING-3.7.md file to clarify that, when adding the following code to the Capfile:
```
require "capistrano/scm/git"
install_plugin Capistrano::SCM::Git
```
it should come after `require 'capistrano/deploy'`.

As reported in [Stack Overflow](http://stackoverflow.com/questions/41307044/capistrano-dont-know-how-to-build-task-deploynew-release-path) and in an [issue](https://github.com/capistrano/capistrano/issues/1824).

### DIFF
--- a/UPGRADING-3.7.md
+++ b/UPGRADING-3.7.md
@@ -12,7 +12,7 @@ set :scm, :svn
 To avoid deprecation warnings:
 
 1. Remove `set :scm, ...` from your Capistrano configuration.
-2. Add *one* of the following SCM declarations to your `Capfile`:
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for the PR! Could you please edit the PR description so that it is more than just an unmodified version of the PR template? I.e. explain what this PR is about, and reference any related issue numbers.

mattbrictson:
Looks great! Thanks 🙇

PR #1839, opened 2017-02-03
Show 'airbrussh' is now format variable default

ajbonner:
Format defaults changed from 'pretty' to 'airbrussh' in 3.5. Updating docs to reflect change.


mattbrictson:
Ah, good catch. Thanks! 🙇

ajbonner:
No worries, was just revisiting some old configs and was getting confused until i rechecked the source. Thanks for all your hard work on this project!

PR #1840, opened 2017-02-04
Git plugin: shellescape git_wrapper_path

olleolleolle:
### Summary

Fixes a bug where deploy would fail if `:local_user` contained a paranthesis; the produced git-ssh suffix is now shell-escaped. See #1742.

(The Git plugin's setting `git_wrapper_path` is a concatenated string, based on three parts. One of them is the username, which can include any character.)

### Short checklist

### Other Information

The previous behavior stopped the rubygems.org staging server from getting a deploy done after upgrading to this newer version of Capistrano (which has the `git_wrapper_path` way of wrapping Git).

The `:local_user` username seems to have been `(GitHub Web Flow) via ShipIt`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reverse Chronological Order:
 https://github.com/capistrano/capistrano/compare/v3.7.2...HEAD
 
 * [#1835](https://github.com/capistrano/capistrano/pull/1835): Stopped printing parenthesis in ask prompt if no default or nil was passed as argument [(@chamini2)](https://github.com/chamini2)
+* [#1840](https://github.com/capistrano/capistrano/pull/1840): Git plugin: shellescape git_wrapper_path [(@olleolleolle)](https://github.com/olleolleolle)
 * Your contribution here!
 
 ## `3.7.2` (2017-01-27)
--- a/lib/capistrano/scm/git.rb
+++ b/lib/capistrano/scm/git.rb
@@ -1,4 +1,5 @@
 require "capistrano/scm/plugin"
+require "shellwords"
 
 class Capistrano::SCM::Git < Capistrano::SCM::Plugin
   def set_defaults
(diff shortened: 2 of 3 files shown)


olleolleolle:
A variant fix would be to 

```ruby
require 'shellwords'
suffix = [:application, :stage, :local_user].map { |key| fetch(key).to_s }.join("-").shellescape
```

will-in-wi:
Offhand, I think I prefer the shellescape method, since that puts the onus on the Ruby standard library to keep track of what needs to be escaped.

olleolleolle:
@will-in-wi I changed the PR to use the standard library instead.

will-in-wi:
@mattbrictson @leehambley This makes sense to me.

olleolleolle:
@will-in-wi Now, with a quite simple unit test, too.

mattbrictson:
Looks great. Thanks! ✨

leehambley:
Agreed on preferring Shellescape

Am 04.02.2017 16:19 schrieb "William Johnston" <notifications@github.com>:

olleolleolle:
I'd like to point to my failure over here: https://github.com/rubygems/rubygems.org/pull/1558#event-949587991 Hm.

sonalkr132:
Following is where `git_wrapper_path` gets used:
```Ruby
task :wrapper do
  on release_roles :all do
    execute :mkdir, "-p", File.dirname(fetch(:git_wrapper_path))
    upload! StringIO.new("#!/bin/sh -e\nexec /usr/bin/ssh -o PasswordAuthentication=no -o StrictHostKeyChecking=no \"$@\"\n"), fetch(:git_wrapper_path)
    execute :chmod, "700", fetch(:git_wrapper_path)
  end
end
```

`upload!` probably doesn't know about escape characters and makes file with name `/tmp/git-ssh-rubygems-staging-web-flow\ \(GitHub\ Web\ Flow\)\ via\ Shipit.sh`. When same filename is passed to `chmod` in next line, it tries to find `/tmp/git-ssh-rubygems-staging-web-flow (GitHub Web Flow) via Shipit.sh` and fails.

Related (:question:) : https://github.com/capistrano/sshkit/issues/333

mattbrictson:
@olleolleolle Can you open a new Capistrano GitHub issue so that we can move discussion there? This seems like a legitimate bug and I don't want it to get lost here.

olleolleolle:
@mattbrictson Ta-da!

PR #1841, opened 2017-02-05
Switch Travis from rbx-2 to rbx-3

mattbrictson:


mattbrictson:
I got rbx-3 to work, but it is *really* slow (builds take 5-10x longer). Not worth it.

PR #1843, opened 2017-02-10
Properly shell escape git:wrapper steps

### LINKED ISSUE #1842: Bug: wrapper is uploaded quoted but is already shell-escaped
- `git_wrapper_path` is created shellescaped
- later, in `:wrapper`, the `upload!` invocation uploads the file inside quotes ("") which makes the shellescape _double_ (or literal, if you will)

See #1840, [or the exact comment](https://github.com/capistrano/capistrano/pull/1840#issuecomment-277591323)

---

### Steps to reproduce

Have a UNIX username with escape-worthy characters in, e.g: `Charlemagne (Charles the Great)` (spaces AND parentheses).

This is an unlikely and horrible scenario, but which happened.

### Expected behavior

The `upload!` and `chmod` steps should refer to the same file.

### Actual behavior

The `upload!` and `chmod` steps refer to different files if any escapable characters are in the created filename. This will fail the `chmod` step, which now probably points to an unknown file.

mattbrictson:
The `upload!` method does escaping of its own internally, so we don't need to do anything special there. Likewise environment variables are already handled. The only place we need to explicitly `shellescape` is when executing the `chmod` command.

Verify this all works as expected by changing the user in the Cucumber features to be `(GitHub Web Flow) via ShipIt`. This user is now used when exercising the `git:check` task.

Fixes #1842.

### Short checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ https://github.com/capistrano/capistrano/compare/v3.7.2...HEAD
 
 * [#1835](https://github.com/capistrano/capistrano/pull/1835): Stopped printing parenthesis in ask prompt if no default or nil was passed as argument [(@chamini2)](https://github.com/chamini2)
 * [#1840](https://github.com/capistrano/capistrano/pull/1840): Git plugin: shellescape git_wrapper_path [(@olleolleolle)](https://github.com/olleolleolle)
+* [#1843](https://github.com/capistrano/capistrano/pull/1843): Properly shell escape git:wrapper steps - [@mattbrictson](https://github.com/mattbrictson)
 * Your contribution here!
 
 ## `3.7.2` (2017-01-27)
--- a/features/step_definitions/assertions.rb
+++ b/features/step_definitions/assertions.rb
@@ -1,10 +1,12 @@
+require "shellwords"
+
 Then(/^references in the remote repo are listed$/) do
   expect(@output).to include("refs/heads/master")
 end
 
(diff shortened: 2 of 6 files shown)


mattbrictson:
@olleolleolle Does this fix work for you?

mattbrictson:
I tested a deployment and it worked:

```ruby
# config/deploy.rb
set :local_user, "(Git Web User) via ShipIt"
```

```
$ cap production deploy:check
00:00 git:wrapper
      01 mkdir -p /tmp
    ✔ 01 deployer@IPADDR 2.306s
      Uploading /tmp/git-ssh-myapp-production-(Git Web User) via ShipIt.sh 100.0%
      02 chmod 700 /tmp/git-ssh-myapp-production-\(Git\ Web\ User\)\ via\ ShipIt.sh
    ✔ 02 deployer@IPADDR 0.071s
00:02 git:check
      01 git ls-remote REMOTEREPO.git HEAD
      01 614bd24b0e18581683dd291bf55ea08c87ed4378 HEAD
    ✔ 01 deployer@IPADDR 1.603s
```

So I'll consider this good to merge.

sonalkr132:
Can you please release a version?
giph1y

mattbrictson:
Capistrano is generally released every 6 weeks. So the next rubygems release will be around March 10th, give or take. In the meantime you can point your Gemfile at the master branch, like this:

```ruby
gem "capistrano", :github => "capistrano/capistrano"
```

PR #1845, opened 2017-02-11
Update documentation styles: add bottom margin to code blocks.

harrykiselev:
Hello!

There is a small update for your documentation styles.

The update fixes annoying small margins (more precisely a lack of it) between block of code and folowed after that paragraph of text.

**After applying changes from my commit we got that (first half of image is before changes/ next half - after changes):**

pull-request-image


mattbrictson:
Nice improvement ✨ 
Thanks for your contribution!

Issue #1427, opened 2015-05-04
Setting user per-host isn't working


maxim:

leehambley:
It likely has to do with the changes to server resolution. One workaround would be to specify the username and domain name in a global (`~/.ssh/config`) or a project local `~/yourproh/.ssh/config`. See https://github.com/capistrano/sshkit/blob/5790401bf5a7f9bdbf2d9ff6f5867001ce465c66/CHANGELOG.md#120, I thought it was documented somewhere else, now I see it's only in the changelog, no wonder nobody knows about it :)

It's a shitty workaround, I know, but this way the hosts will look totally different to sshkit:

```
Host 1
   User userone
   Host hostone

Host 2
   User usertwo
   Host hostone
```

Then connect to `1` and `2` (choose better names!) in SSHKit.

leehambley:
You might also try:

``` ruby
server 'deployer@foo.example.org', roles: [:full]
```

maxim:
@leehambley thanks for alternatives. Your latter `'deployer@…'` version didn't help, but I realized that if I comment out the 2nd declaration in my original example like this:

```ruby
server 'foo.example.org', user: 'deployer', roles: [:util]
# role :util_exec, 'foo.example.org', user: 'runner', no_release: true
```

then `whoami` actually shows deployer as expected. So having 1 server in 2 roles with different user assigned to it is no longer possible this way (e.g. for deployment tasks you need deployer user but for running console, scripts, etc you need runner user on the same host). You're saying .ssh/config is the workaround? I agree it's far from ideal, at least because you now gotta specify your servers in multiple places. Any plans to restore the old behavior in this case?

leehambley:
The server matching code has undergone quite some changes, you can check the code to see who the "Blame" points to and bring the discussion back here. If you can write a test case for your specific problem using our suite (there's plenty of reference code) then we have a concrete place to start. 

It seems however that whatever we do with "merge the servers intelligently" vs. "last value wins" vs. "allow multiple" we'll always get it wrong for some people.

I don't know that I agree that the ssh config is a bad workaround, I think the following is kinda neat:

``` 
# ssh config in ./myproject/.ssh/config
Host util_server
    Host foo.example.org
    User runner
```

paired with a single line change:

``` diff
- role :util_exec, 'foo.example.org', user: 'runner', no_release: true
+ role :util_exec, util_server, no_release: true
```

is all that bad?

leehambley:
[...]

PR #1846, opened 2017-02-12
Make add_role O(hosts) & fix port handling

dbenamy:
### Summary

This makes add_role O(hosts) instead of O(hosts^2). I'm using cap in a deployment with enough servers that the previous behavior makes it take 7+ seconds to start.

I also think I fixed a bug in add_host's handling of ports, but I'm not totally sure of the intended behavior so let me know if I misunderstood!

Each commit stands alone if that helps with reviewing.

### Other Information

I benchmarked using

```
from subprocess import check_output
from time import time

for cnt in range(0, 12001, 100):
    with file('Capfile', 'w') as f:
        f.write("require 'capistrano/setup'\n")
        f.write("role :r, %%w(%s)\n" % ' '.join(['i-%017d' % i for i in range(cnt)]))
    start = time()
    check_output(['bin/cap', '--rakefile=Capfile', '-T'])
    print '%d,%f' % (cnt, time() - start)
```

`RUBYLIB=lib python bench.py`

screen shot 2017-02-12 at 1 23 50 am

The detailed data is at https://docs.google.com/spreadsheets/d/14OvbOwF8eGflgWhYleI9RDkpmfUtDQDF-iH2oEIFPko. The commits shown there don't match the PR any more because I squashed in a tiny fix. They should be equivalent to 5aea6dc for baseline, and 45e1920 for the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ https://github.com/capistrano/capistrano/compare/v3.7.2...HEAD
 * [#1835](https://github.com/capistrano/capistrano/pull/1835): Stopped printing parenthesis in ask prompt if no default or nil was passed as argument [(@chamini2)](https://github.com/chamini2)
 * [#1840](https://github.com/capistrano/capistrano/pull/1840): Git plugin: shellescape git_wrapper_path [(@olleolleolle)](https://github.com/olleolleolle)
 * [#1843](https://github.com/capistrano/capistrano/pull/1843): Properly shell escape git:wrapper steps - [@mattbrictson](https://github.com/mattbrictson)
+* [#1846](https://github.com/capistrano/capistrano/pull/1846): Defining a role is now O(hosts) instead of O(hosts^2) [(@dbenamy)](https://github.com/dbenamy)
+* [#1846](https://github.com/capistrano/capistrano/pull/1846): add_host will add a new host in a case where it used to incorrectly update an existing one (potentially breaking) [(@dbenamy)](https://github.com/dbenamy)
 * Your contribution here!
 
 ## `3.7.2` (2017-01-27)
--- a/lib/capistrano/configuration/server.rb
+++ b/lib/capistrano/configuration/server.rb
@@ -64,6 +64,7 @@ module Capistrano
       end
 
       def matches?(other)
+        # This matching logic must stay in sync with `Servers#add_host`.
         hostname == other.hostname && port == other.port
       end
 
--- a/lib/capistrano/configuration/servers.rb
+++ b/lib/capistrano/configuration/servers.rb
@@ -10,11 +10,14 @@ module Capistrano
       def add_host(host, properties={})
         new_host = Server[host]
         new_host.port = properties[:port] if properties.key?(:port)
-        if (server = servers.find { |s| s.matches? new_host })
-          server.user = new_host.user if new_host.user
-          server.with(properties)
+        # This matching logic must stay in sync with `Server#matches?`.
+        key = ServerKey.new(new_host.hostname, new_host.port)
+        existing = servers_by_key[key]
+        if existing
+          existing.user = new_host.user if new_host.user
+          existing.with(properties)
         else
-          servers << new_host.with(properties)
+          servers_by_key[key] = new_host.with(properties)
         end
       end
 
@@ -27,7 +30,7 @@ module Capistrano
 
       def roles_for(names)
         options = extract_options(names)
-        s = Filter.new(:role, names).filter(servers)
+        s = Filter.new(:role, names).filter(servers_by_key.values)
         s.select { |server| server.select?(options) }
       end
 
@@ -54,13 +57,15 @@ module Capistrano
       end
 
       def each
-        servers.each { |server| yield server }
+        servers_by_key.values.each { |server| yield server }
       end
 
       private
 
-      def servers
-        @servers ||= []
+      ServerKey = Struct.new(:hostname, :port)
+
+      def servers_by_key
+        @servers_by_key ||= {}
       end
 
       def extract_options(array)
--- a/lib/capistrano/configuration/servers.rb
+++ b/lib/capistrano/configuration/servers.rb
@@ -9,6 +9,7 @@ module Capistrano
 
       def add_host(host, properties={})
         new_host = Server[host]
+        new_host.port = properties[:port] if properties.key?(:port)
         if (server = servers.find { |s| s.matches? new_host })
           server.user = new_host.user if new_host.user
           server.with(properties)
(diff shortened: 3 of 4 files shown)


mattbrictson:
Thanks for the PR and the detailed supporting data!

I may not have time to review the code in detail right away, but in the meantime I am curious to learn more about the potential breaking change. Does your change perhaps overlap with the breaking changes proposed in #1427?

mattbrictson:
@leehambley or @will-in-wi could you take the lead on reviewing this one?

leehambley:
Thanks for the PR @dbenamy , doubly so for including some plots about performance, it's a rare luxury in our position as maintainers to receive such things!

I'd love to know (just for fun) *what* exactly you're doing that warrants caring about the big-O complexity of some mundane part of setting up the roles, I might ping you off-list, whatever you're doing, sounds like a serious case!

As best [I can see](https://github.com/capistrano/capistrano/pull/1846/files#diff-0d9345682ea186c858c7b2c5f1fbceb9R68) we're still not considering the username as part of the host definition. 

Honestly, having re-read #1427 I can see both cases being valid, but we do just have to decide one. I consider #1427 to be somewhat moot since someone can commit a `./ssh/config` file with usernames and use descriptive hostnames, which is how I personally use Cap, but that doesn't need to affect this PR. Concretely: I think it's ok for a "server" to be the combo of hostname/ip and port.

I see that you changed [the spec](https://github.com/dbenamy/capistrano/blob/629e26284cb50b4980dc9d2a145cd457d3ea77bc/spec/lib/capistrano/configuration/servers_spec.rb#L131-L140) for how we treat usernames and ports in relation to hostnames, I believe this is the "potential breaking change" that @mattbrictson is talking about:

``` ruby
        it "can accept multiple servers with the same hostname but different ports or users" do
          servers.add_host("1", roles: [:app, "web"], test: :value, port: 12)
          expect(servers.count).to eq(2)
          servers.add_host("1", roles: [:app, "web"], test: :value, port: 34)
          servers.add_host("1", roles: [:app, "web"], test: :value, user: "root")
          servers.add_host("1", roles: [:app, "web"], test: :value, user: "deployer")
          servers.add_host("1", roles: [:app, "web"], test: :value, user: "root", port: 34)
          servers.add_host("1", roles: [:app, "web"], test: :value, user: "deployer", port: 34)
          servers.add_host("1", roles: [:app, "web"], test: :value, user: "deployer", port: 56)
          expect(servers.count).to eq(4)
        end
```

I always struggle parsing such banal code, but I suspect what you're expecting is the following status at the end of this block:

```
#<Capistrano::Configuration::Server:0x007fccbb6af8a8 @keys=[], @local=false, @user="deployer", @hostname="1", @port=nil, @properties=#<Capistrano::Configuration::Server::Properties:0x007fccbb6af3f8 @properties={:test=>:value}, @roles=#<Set: {:app, :web}>>>
#<Capistrano::Configuration::Server:0x007fccbb6aeef8 @keys=[], @local=false, @user=nil,        @hostname="1", @port=12,  @properties=#<Capistrano::Configuration::Server::Properties:0x007fccbb6aeac0 @properties={:test=>:value}, @roles=#<Set: {:app, :web}>>>
#<Capistrano::Configuration::Server:0x007fccbb6ae480 @keys=[], @local=false, @user="deployer", @hostname="1", @port=34,  @properties=#<Capistrano::Configuration::Server::Properties:0x007fccbb6ae048 @properties={:test=>:value}, @roles=#<Set: {:app, :web}>>>
#<Capistrano::Configuration::Server:0x007fccbe07b4e8 @keys=[], @local=false, @user="deployer", @hostname="1", @port=56,  @properties=#<Capistrano::Configuration::Server::Properties:0x007fccbe07b088 @properties={:test=>:value}, @roles=#<Set: {:app, :web}>>>
```

According to these things we've got four server, all with the same hostname, but with different ports, I could live with that as a small change in the docu, and clean up the examples above and explain how it works, and why, and how the "workaround" should be employed?

dbenamy:
Thanks for taking a look!

I hadn't seen https://github.com/capistrano/capistrano/issues/1427 before. I didn't go into this looking to change any behavior; I just noticed while working on it that there was an inconsistency in some cases where adding a server with the port specified using the properties hash updates a server
```
servers.add_host('host1')
=> [#<Capistrano::Configuration::Server:0x007f95130d1e98 @keys=[], @local=false, @user=nil, @hostname="host1", @port=nil>]
servers.add_host('host1', port: 1234)
=> #<Capistrano::Configuration::Server:0x007f95130d1e98 @keys=[], @local=false, @user=nil, @hostname="host1", @port=1234>
```
while adding it with the port specified in the host string creates a second
```
servers.add_host('host1')
=> [#<Capistrano::Configuration::Server:0x007f81c51f2860 @keys=[], @local=false, @user=nil, @hostname="host1", @port=nil>]
servers.add_host('host1:1234')
=> [#<Capistrano::Configuration::Server:0x007f81c51f2860 @keys=[], @local=false, @user=nil, @hostname="host1", @port=nil>, #<Capistrano::Configuration::Server:0x007f81c51ca720 @keys=[], @local=false, @user=nil, @hostname="host1", @port=1234>]
```

I listed is as a potential breaking change in the CHANGELOG since it's possible that people are relying on the existing behavior.

> According to these things we've got four server, all with the same hostname, but with different ports, I could live with that as a small change in the docu, and clean up the examples above and explain how it works, and why, and how the "workaround" should be employed?

Cool. To be clear- I don't have a ton of time to get involved in what seems to be a very long running discussion about how the semantics should work. If this port handling fix is a clear win, I'm happy to update whatever needs updating, if you could point me at exactly what that is. If it makes accepting this PR any harder though, I can re-implement the previous behavior with the O(n) improvement. Just let me know!

> I'd love to know (just for fun) what exactly you're doing that warrants caring about the big-O complexity of some mundane part of setting up the roles, I might ping you off-list, whatever you're doing, sounds like a serious case!

I'm using cap in a deployment with enough servers that this is relevant! :-) There are maybe 30 roles, some with hundreds of hosts, and it takes 7 - 8 seconds for cap to start up before it does any actual tasks.

leehambley:
> Cool. To be clear- I don't have a ton of time to get involved in what seems to be a very long running discussion about how the semantics should work.

Agreed!

I think we've done enough here. I'd take it on my plate to open another ticket to fix the documentation to explain how it is now, and provide some examples.

I want to avoid bikeshedding around some trivialities that you happened to fall over, when I don't think any of us truly feel like planting the proverbial flag and proclaiming the one true way to name or define servers.

Other maintainers, @mattbrictson and @will-in-wi any stake in this? First pass over the review LGTM although I had some weird local test failures that travis didn't reproduce, I think we're fine ?

mattbrictson:
LGTM 👍
Feel free to merge, @leehambley

dbenamy:
Thanks! Any idea when the next release will be?

mattbrictson:
Capistrano is generally released every 6 weeks. So the next rubygems release will be around March 10th, give or take.

dbenamy:
Excellent, thanks again!

PR #1847, opened 2017-02-12
Update style for alert messages at documentation pages

harrykiselev:
### Summary

This PR is attempt to fix alert messages style at capistrano documentation.

### Other Information

Hello!

At first I found forgotten class (.alert-box) at alert message on [Installation page](http://capistranorb.com/documentation/getting-started/installation/).

This page looked like:

After I added this class, alert message began to look like this (as all alert messages at documentation):

It looks like alerts lost their line-height property (compared with paragraphs of text on same page).

To fix this I did not touch documentation stylesheets (.alert-box class styles). I took advantage of the fact that «p» tags has this property and have changed tag for all alerts from «div» to «p».

After that alert messages began looks like this (before and after update versions):

It touched some list of pages:

* [Installation](http://capistranorb.com/documentation/getting-started/installation/)
* [Preparing Your Application](http://capistranorb.com/documentation/getting-started/preparing-your-application/)
* [How can I set Capistrano configuration paths?](http://capistranorb.com/documentation/faq/how-can-i-set-capistrano-configuration-paths/)

It seems to me that the styles of alert messages look more pleasant.

Thanks :)

### DIFF
--- a/docs/css/capistrano.css
+++ b/docs/css/capistrano.css
@@ -95,6 +95,30 @@ h1, h2, h3, h4, h5, h6 {
   margin-bottom: 1.25em;
 }
 
+.alert-box {
+  font-weight: normal;
+  color: black;
+
+  background-color: #d7ecfa;
+  border: none;
+  padding-left: 20px;
+}
+
+.alert-box.alert {
+  background-color: #f7e4e1;
+  color: black;
+}
+
+.alert-box.success {
+  background-color: #e1faea;
+  color: black;
+}
+
+.alert-box.secondary {
+  background-color: #eaeaea;
+  color: black;
+}
+
 /*p code, li code {
   padding: 3px;
   background-color: #E6E6E6;
--- a/docs/documentation/faq/how-can-i-set-capistrano-configuration-paths/index.markdown
+++ b/docs/documentation/faq/how-can-i-set-capistrano-configuration-paths/index.markdown
@@ -16,7 +16,7 @@ set :stage_config_path, 'cap/stages'
 # previous variables MUST be set before 'capistrano/setup'
 require 'capistrano/setup'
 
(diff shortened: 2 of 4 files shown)


mattbrictson:
Hi, thanks for taking the time to improve the styles!

It looks like the `.alert-box` has not been styled for a long time, so I don't know what the original intent was for the visual effect.

I think the style you've added in this PR makes the alerts a bit hard to read, in particular the blue and white text on the solid red background. Could you change it to something a bit more subtle? For example, I like how the [Rails Guides](http://guides.rubyonrails.org/getting_started.html) use a yellow box that calls attention to the text without being overbearing.

screen shot 2017-02-14 at 6 54 15 pm

harrykiselev:
Matt, sorry for a long pause between messages :)

All documentation styles based on one of the older version of Foundation front-end framework.
Colors of alerts messages was chosen by creators of the framework.

Colors in version that used in capistrano docs looks like this (image from docs):


In up-to-date version of framework colors was changed and looks like:


I think so best (simply and clearly) solution is to update colors, remove border and set text color to black and font-weight to normal by default. Updated version of the alerts looks more subtle and not overbearing. :)

What do you think about it? :)

leehambley:
> I think so best (simply and clearly) solution is to update colors, remove border and set text color to black and font-weight to normal by default. Updated version of the alerts looks more subtle and not overbearing. :)

Sounds good, I was 99% likely responsible for those decisions, they look must neater now.

mattbrictson:
Looks great. Thanks again for your contribution! 🙇

leehambley:
chic, thank you ! 👍 

screenshot 2017-02-24 17 50 59

harrykiselev:
@mattbrictson, @leehambley I am very glad that you liked, thank you! :)

PR #1853, opened 2017-02-19
Explain the six-week release cadence

mattbrictson:
I've mentioned that I want Capistrano to use a six-week release cadence. Any objections to making this "official policy" by mentioning it in the changelog?

I chose six weeks because I think it strikes a good balance:

* Often enough that contributions are regularly published
* Not too often that users are overwhelmed trying to keep up with new versions

I also prefer a regular schedule (as opposed to "release whenever it's ready") because:

* Contributors know when their code will be published
* It encourages stability in `master` (since a release is always imminent)
* A steady project "heartbeat" keeps the community interested and involved

/cc @will-in-wi @leehambley

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Capistrano 3.x Changelog
 
-Reverse Chronological Order:
+All notable changes to this project will be documented in this file, in reverse chronological order. This project adheres to [Semantic Versioning](http://semver.org).
+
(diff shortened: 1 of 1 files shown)


leehambley:
No objections, I like the idea and think more projects should do it.

Am 19.02.2017 23:22 schrieb "Matt Brictson" <notifications@github.com>:

will-in-wi:
This makes a ton of sense to me, and seems to be a good balance of not too much work, and getting fixes out in a timely manner.

PR #1854, opened 2017-02-24
Drop dependency on `capistrano-harrow' gem

leehambley:
Related to https://github.com/harrowio/capistrano-harrow/issues/4 - easy fix, just sending as a PR to get the other maintainers in the loop.


mattbrictson:
We also need to remove these lines, right?

https://github.com/capistrano/capistrano/blob/cb54892912340a89f4e87da3df094e44ccd93205/lib/Capfile#L5-L7

```ruby
require "capistrano/harrow"
require "capistrano/harrow/plugin"
install_plugin Capistrano::Harrow::Plugin
```

leehambley:
Done, and force pushed (because in a branch, thanks!)

mattbrictson:
Looks good. I'll run the feature tests locally and merge once that's done.

PR #1855, opened 2017-02-24
Remove nagging popover from docs site

leehambley:
Another nag removed, for the better!


mattbrictson:
Looks good. Thanks!

Does it make sense to remove the `.harrowBanner` styles as well?
https://github.com/capistrano/capistrano/blob/8e51b227d4639bfde65ff173dd17bc74a4d1aa84/docs/css/capistrano.css#L29-L86

leehambley:
Indeed, I did think about keeping it incase we need a similar banner, but someone can dig up the history in Git if they are so inclined.

mattbrictson:
Nice 👍

PR #1856, opened 2017-02-25
Fix hg repo_tree implementation

### LINKED ISSUE #1849: :repo_tree configuration doesn't work with Mercurial
Hi,

I'm using Capistrano 3.7.1 and I wanted to use the `repo_tree` option.
Unfortunately with Mercurial it doesn't work, and deployment fails.

```
00:02 hg:create_release
      01 hg archive --type tgz -p . -I website --rev develop | tar -x --strip-components 1 -f - -C /var/www/html/releases/20170217155548
      01 hg archive: invalid arguments
      01 tar: This does not look like a tar archive
      01 tar: Skipping to next header
      01 tar: Exiting with failure status due to previous errors
```

`git archive` writes to the standard output, but `hg archive` doesn't.

mattbrictson:
`hg archive` does not write to standard out, so Capistrano's `git` implementation of the `:repo_tree` option does not work for `hg`. As a workaround, generate a temporary tar file, extract it, and then remove it. This is not as efficient or elegant as doing it via stdout and pipes in a single line, but it should work.

Fixes #1849.

@zebastiane can you test the branch to see if it works? Do you have any concerns about the implementation? Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ https://github.com/capistrano/capistrano/compare/v3.7.2...HEAD
 * [#1846](https://github.com/capistrano/capistrano/pull/1846): add_host will add a new host in a case where it used to incorrectly update an existing one (potentially breaking) [(@dbenamy)](https://github.com/dbenamy)
 * [capistrano-harrow#4](https://github.com/harrowio/capistrano-harrow/issues/4): Drop dependency on `capistrano-harrow` gem. Gem can still be installed separately [(@leehambley)](https://github.com/leehambley)
 * Run `svn switch` to work with svn branches if repo_url is changed
+* [#1856](https://github.com/capistrano/capistrano/pull/1856): Fix hg repo_tree implementation - [@mattbrictson](https://github.com/mattbrictson)
 * Your contribution here!
 
 ## `3.7.2` (2017-01-27)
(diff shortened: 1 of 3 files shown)


zebastiane:
Hi,

I tested but unfortunately it doesn't work.
The archive can't be extracted because the directory doesn't exists (how does `git` handle this?).

```
00:03 hg:create_release
      01 hg archive -p . -I htdocs --rev develop/4.0.0 /tmp/a3d6575bc38316c79133.tar
      02 tar -x --strip-components 1 -f /tmp/a3d6575bc38316c79133.tar -C /var/www/html/releases/20170302141707
      02 tar: /var/www/html/releases/20170302141707: Cannot open: No such file or directory
      02 tar: Error is not recoverable: exiting now
```

The archive content is good.

mattbrictson:
Hmm. Interesting. Thanks for checking! I am not sure why it git's case `tar` is able to create the directory. I'll look into it further.

mattbrictson:
Ah, `git.rake` runs `execute :mkdir, "-p", release_path` first. Should be an easy fix.

mattbrictson:
@zebastiane I've force-pushed a fix. Can you try it again?

zebastiane:
@mattbrictson It works!

mattbrictson:
Great! Thanks for testing. I'll merge it.

PR #1859, opened 2017-02-26
Move git-specific repo_url logic into git plugin

### LINKED ISSUE #1715: Refactor `repo_url` to remove Git-specific behavior
Right now the implementation of the `repo_url` helper method has some special logic that is Git-specific. It looks for the presence of `:git_http_username` and `:git_http_password` and uses these to mutate the original `:repo_url`.

From `dsl/paths.rb`:

``` ruby
def repo_url
  require "cgi"
  require "uri"
  if fetch(:git_http_username) && fetch(:git_http_password)
    URI.parse(fetch(:repo_url)).tap do |repo_uri|
      repo_uri.user     = fetch(:git_http_username)
      repo_uri.password = CGI.escape(fetch(:git_http_password))
    end.to_s
  elsif fetch(:git_http_username)
    URI.parse(fetch(:repo_url)).tap do |repo_uri|
      repo_uri.user = fetch(:git_http_username)
    end.to_s
  else
    fetch(:repo_url)
  end
end
```

This is a handy feature, but I believe this is the wrong place to implement it. It should live in the Git SCM plugin.

Since we eventually want to move all SCM implementations out of core and into separate plugins, this code will need to be moved as well. I suggest we do this after #1572 is merged.

mattbrictson:
The implementation of the `repo_url` helper method had some special logic that was Git-specific. It looked for the presence of `:git_http_username` and `:git_http_password` and used these to mutate the original `:repo_url`.

Since we eventually want to move all SCM implementations out of core and into separate plugins, I have moved this code into the Git plugin.

Also add a unit test (this feature was previously untested).

Fixes #1715.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ https://github.com/capistrano/capistrano/compare/v3.7.2...HEAD
 * [#1856](https://github.com/capistrano/capistrano/pull/1856): Fix hg repo_tree implementation - [@mattbrictson](https://github.com/mattbrictson)
 * [#1857](https://github.com/capistrano/capistrano/pull/1857): Don't emit doctor warning when repo_tree is set - [@mattbrictson](https://github.com/mattbrictson)
 * [#1860](https://github.com/capistrano/capistrano/pull/1860): Allow cap to be run within subdir and still work - [@mattbrictson](https://github.com/mattbrictson)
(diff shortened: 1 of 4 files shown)


mattbrictson:
@leehambley @will-in-wi any strong opinions on this PR or #1858? I'll probably merge them next week.

leehambley:
LGTM!

PR #1860, opened 2017-02-26
Allow cap to be run within subdir and still work

### LINKED ISSUE #1832: Allow cap command to work from within subdirectory
Using capistrano 3.7.1, if I run `cap qa deploy` from a subfolder, it says

    Stage not set, please call something such as `cap production deploy`, where production is a stage you have defined.

So my `cap qa deploy` has turned into `cd .. && cap qa deploy && cd -`

It's a very minor annoyance but I run into it all the time.  Would it be okay for cap to behave like rake and git in that it searches parent dirs for its files?

mattbrictson:
Much like rake can find its Rakefile even when run in a subdirectory, cap can now find its Capfile. Since cap is built on top of rake, this was really just restoring default behavior that was accidentally overridden.

Also add a feature test to ensure it works.

Fixes #1832.


mattbrictson:
@dan-passaro Can you test whether this branch fixes #1832 for you?

dan-passaro:
Yes @mattbrictson, this fixes #1832 for my projects, thank you!!

PR #1875, opened 2017-04-09
Fix build failures due to the release of RuboCop 0.48.1

mattbrictson:
Travis CI for `master` (and all PRs) is currently broken because `Lint/AmbiguousBlockAssociation` in RuboCop 0.48.1 forbids this construct:

```ruby
set :foo, -> { "bar" }
```

It apparently wants:

```ruby
set(:foo, -> { "bar" })
```

This recommendation isn't desirable in Capistrano, so I disabled it.

I also pinned our RuboCop dependency so that new versions won't suddenly cause our builds to break in the future. We can always manually upgrade RuboCop if desired.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -3,6 +3,9 @@ AllCops:
   DisplayStyleGuide: true
   TargetRubyVersion: 2.0
 
+Lint/AmbiguousBlockAssociation:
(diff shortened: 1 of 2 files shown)


will-in-wi:
This makes sense to me.

PR #1883, opened 2017-04-26
Skip invalid releases instead of skipping the whole cleanup

dhiemstra:
The change introduced in #1811 skips the whole cleanup processes if an invalid folder is found.
In our case we create an `initial` folder in the releases during the initial server configuration so we can have an existing `current` symlink so nginx wont crash.

After upgrading capistrano no releases were cleaned anymore and just a warning was given which is barely noticeable by the big loads of capistrano output. This caused our production server disk to run out of free space.

I suggest that the cleanup process should still happen and the invalid folders should be skipped.

### DIFF
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -149,9 +149,12 @@ namespace :deploy do
   task :cleanup do
     on release_roles :all do |host|
       releases = capture(:ls, "-x", releases_path).split
-      if !(releases.all? { |e| /^\d{14}$/ =~ e })
-        warn t(:skip_cleanup, host: host.to_s)
-      elsif releases.count >= fetch(:keep_releases)
+      invalid_releases = releases.select { |e| /^\d{14}$/ =~ e }
+      if invalid_releases.any?
+        warn t(:invalid_releases_found, host: host.to_s)
+        releases -= invalid_releases
+      end
+      if releases.count >= fetch(:keep_releases)
         info t(:keeping_releases, host: host.to_s, keep_releases: fetch(:keep_releases), releases: releases.count)
         directories = (releases - releases.last(fetch(:keep_releases)))
         if directories.any?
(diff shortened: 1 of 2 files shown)


mattbrictson:
Hi, thanks for the PR!

I like the idea of this change, but for the PR to be seriously considered, it needs:

* A passing Travis CI build ([see failure](https://travis-ci.org/capistrano/capistrano/builds/225968887))
* Tests ([as requested above](https://github.com/capistrano/capistrano/pull/1883#issuecomment-297369751))
* A changelog entry ([as requested above](https://github.com/capistrano/capistrano/pull/1883#issuecomment-297369751))

If you need any help providing these, let me know and I will do my best to assist. Thanks!

lib/capistrano/tasks/deploy.rake
```diff
@@ -149,9 +149,12 @@ namespace :deploy do
   task :cleanup do
     on release_roles :all do |host|
       releases = capture(:ls, "-x", releases_path).split
-      if !(releases.all? { |e| /^\d{14}$/ =~ e })
-        warn t(:skip_cleanup, host: host.to_s)
-      elsif releases.count >= fetch(:keep_releases)
+      invalid_releases = releases.select { |e| /^\d{14}$/ =~ e }
```
darwin67: this should be `reject` instead. you're taking out the actual releases here and it will just result in not rotation again.

darwin67:
@dhiemstra Hi, any update on this PR. I'm encountering this issue too and it would be great to see this change get it.

mattbrictson:
@darwin67 I'm not sure the original author is still available to see this one through. If you'd like to create a PR of your own, please do!

darwin67:
@mattbrictson thanks, i'll do that

darwin67:
i think this PR can be closed now since #1899 was merged

mattbrictson:
Closed in favor of #1899.

Thanks @darwin67!

PR #1887, opened 2017-05-12
`git ls-remote` should be performed within repo_path

yskkin:
### Summary

Checking reachability of git repo should be done inside git repo in order to utilize local git config.

For example, AWS CodeCommit [recommends](http://docs.aws.amazon.com/codecommit/latest/userguide/setting-up-https-unixes.html#setting-up-https-unixes-credential-helper) setting up credential helper.
If config is saved locally, current `git:check` task fails.

### DIFF
--- a/lib/capistrano/scm/tasks/git.rake
+++ b/lib/capistrano/scm/tasks/git.rake
@@ -15,8 +15,10 @@ namespace :git do
   task check: :'git:wrapper' do
     fetch(:branch)
     on release_roles :all do
-      with fetch(:git_environmental_variables) do
-        git_plugin.check_repo_is_reachable
+      within repo_path do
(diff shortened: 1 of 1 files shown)


mattbrictson:
Hi, thanks for the PR!

Unfortunately I can't accept this PR as-is because it will not work for an initial deploy: `git:check` is run before the repository has even been cloned. So the local repo will not exist and thus `within repo_path` will fail.

However I don't think this change is actually necessary based on [AWS recommendation](http://docs.aws.amazon.com/codecommit/latest/userguide/setting-up-https-unixes.html#setting-up-https-unixes-credential-helper) that was mentioned. According to that doc, it seems like the suggestion is to use the `--global` configuration for credentials. If the configuration is global, then the working directory should not matter.

PR #1888, opened 2017-05-24
Do not claim to follow SemVer

mvz:
Capistrano introduces breaking changes in minor versions. SemVer requires an update of the major version for such changes. Hence, Capistrano should not claim to follow SemVer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Capistrano 3.x Changelog
 
-All notable changes to this project will be documented in this file, in reverse chronological order. This project adheres to [Semantic Versioning](http://semver.org).
+All notable changes to this project will be documented in this file, in reverse chronological order.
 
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for the PR. Agreed that we do not strictly follow SemVer. But we do care about our users and pay close attention to versioning! I'd rather replace the SemVer sentence with a better explanation of our version policy.

I think we are in pretty close alignment with [what Rails uses](https://github.com/rails/rails/blob/3a628b9338c436001c00a9865b22036abd44a2bd/guides/source/maintenance_policy.md). They say:

```md
Rails follows a shifted version of [semver](http://semver.org/):

**Patch `Z`**

Only bug fixes, no API changes, no new features.
Except as necessary for security fixes.

**Minor `Y`**

New features, may contain API changes (Serve as major versions of Semver).
Breaking changes are paired with deprecation notices in the previous minor
or major release.

**Major `X`**

New features, will likely contain API changes. The difference between Rails'
minor and major releases is the magnitude of breaking changes, and usually
reserved for special occasions.
```

Would you agree this is accurate?

mvz:
> Would you agree this is accurate for Capistrano?

Yes, that seems about right.

mattbrictson:
Closing in favor of #1894.

PR #1890, opened 2017-05-25
Feature/configurable releases and shared folder name

ShoobyBan:
### Summary

Making 'releases' and 'shared' directory names configurable from deployment target.


will-in-wi:
@ShoobyBan: Are you able to address @mattbrictson's concerns in your PR?

ShoobyBan:
I'm trying hard to learn rspec, (honestly, my first attempt) tried something like this but somehow the code seems to run on default value instead of the expects return value.

```ruby
  describe "#releases_path" do
    subject { dsl.releases_path }

    context "with releases custom directory" do
      before do
        paths.expects(:fetch).with(:releases_directory).returns("test123")
      end

      it "returns the releases path with the custom directory" do
        expect(subject.to_s).to eq "/var/www/test123"
      end
    end

  end
```

Any help appreciated.

mattbrictson:
The way I usually write stubs in rspec is like this:

```ruby
allow(paths).to receive(:fetch).with(:releases_directory).and_return("test123")
```

Does that make a difference?

Perhaps the other issue is that the subject should be:

```ruby
subject { paths.releases_path }
```

ShoobyBan:
I've rebased it, added basic tests and included it in the getting started / configuration page. 
Many thanks for your patience and help @mattbrictson

mattbrictson:
Thanks for following up! It looks like something went wrong with the rebase; the PR diff is much bigger than it should be. I'll attempt to resolve the problem move your commits to a new PR.

mattbrictson:
Closed in favor of #1943

PR #1894, opened 2017-06-02
Clarify how Capistrano does its versioning

mattbrictson:
This is a small documentation change to clarify that Capistrano does *not* strictly follow SemVer, as pointed out in #1888. Notably, minor version bumps may occasionally contain small breaking changes.


mattbrictson:
@leehambley @will-in-wi any objections?

leehambley:
None whatsoever @mattbrictson - looks good, we don't break things *that* often though, do we ?

mattbrictson:
"Occasionally" is definitely open for interpretation. We try not to. 😃

will-in-wi:
This makes sense to me too.

Issue #1893, opened 2017-06-01
Capistrano leaks local user account name to deployment log


rbq:
### Steps to reproduce
1. Set up Capistrano project
2. Deploy to remote server

### Expected behavior
* Private information about developer machine config should not leak to remote server
* Local account names that make no sense to others should no be (ab-)used to distinguish between people on a project-wide scale.
* Deployment logs should contain information that is actually useful, such as a human readable, manually configured user names. In many cases that would probably be your Git `user.name` setting for the specific project.

### Actual behavior
* Capistrano leaks local user account name to log without even warning the user about it. This serves little purposes since other parties usually don't know your personal account name. But it might be something that you might not want to share with clients, e.g. a silly Star Wars reference, `admin`, your girlfriends name, some other project or whatever.

mattbrictson:
Hi, thanks for the report! I agree, Capistrano should not surprisingly publish what some might consider personal information.

The `$USER` that you are seeing in the logs is driven by the `:local_user` Capistrano variable. So it is easy to change this to any value you want. For example, you could do this:

```ruby
# In deploy.rb
set :local_user, -> { `git config user.name`.chomp }
```

The bigger issue here is do we want to change the default behavior? If so, how do we do this in a backwards-compatible way?

`:local_user` is used in multiple places:

1. Attribution in `revisions.log`
2. Creating a per-user unique filename for the `:git_wrapper_path`
3. An unknown number of third-party Capistrano plugins may depend on it

The Git `user.name` might work for most projects, but we cannot assume that everyone is using Git. Further, as you point out, `$USER` is not good option either.

Any suggestions?

rbq:
Hi, thanks for the advice about the `:local_user` config key. I think adding (a) this setting to the generated default config and (b) mentioning it in the [documentation](http://capistranorb.com/documentation/getting-started/structure/) would already help a lot in terms of preventing undesirable surprises.

My personal preference for future defaults would be to …

1. decouple the `revisions.log` info from `:local_user`. Introduce a new config key specifically for this purpose (probably something more general like `:deployed_by` that could equally well represent information about the CI server triggering/executing the build)
2. not add any name to `revisions.log` unless `:deployed_by` is `.present?`
3. put a sensible default for `:deployed_by` into newly generated `deploy.rb` files; maybe Git `user.name` `||` unknown
[...]

PR #1895, opened 2017-06-08
Add :local_user setting to config and docs

rbq:
### Summary

This PR adds the `:local_user` setting to the generated `deploy.rb` to encourage users to pick a reasonable setting that identifies their team members and/or CI processes.

It also adds a mention of `:local_user` to the according section of the documentation that deals with the `revisions.log` file.

Partially resolves issue #1893, “Capistrano leaks local user account name to deployment log”.


mattbrictson:
Looks great, thanks! ✨ No need for tests or changelog entry for this PR.

PR #1898, opened 2017-06-16
Explain how security issues should be reported

mattbrictson:
Security issues should not be reported through GitHub; they should go to <security@capistranorb.com>.


leehambley:
Nice, love it.

will-in-wi:
I agree about the responsible disclosure. Although, I'd merge this and take care of that in another ticket.

mattbrictson:
I'm not sure what exactly you both have in mind for the responsible disclosure portion, so I will take @will-in-wi's suggestion and merge this as-is. Please open a follow-up PR when you have a chance!

Issue #547, opened 2013-07-08
Don't run cleanup on rollback by default


dbackeus:
When running `cap:rollback` the `cap:rollback:cleanup` task gets run by default.

However `cap deploy` does not run `cap deploy:cleanup` unless explicitly added as a trigger.

When using unicorn and similar servers that do 0 downtime deploys by pre-loading a new set of server app instances - removing the old deployed directory will cause errors until the new ones have booted.

So optimally `cap:rollback:cleanup` should not run unless explicitly told to do so.

This also adheres to the principle of least surprise as the behaviour becomes consistent with what people are used to from running `cap deploy`.

jsha:
I'd also very much like for `cap:rollback` to not run `cap:rollback:cleanup` by default. I've used Capistrano extensively for deploys at Twitter and for personal projects, and the default `cap:rollback:cleanup` behavior has very frequently gotten in the way of debugging deploy problems. If there's any sort of error during a deploy, usually the evidence of that error is in the release directory that `cap:rollback:cleanup` automatically deletes. This winds up wasting a lot of time trying to track down the problem.

An alternate proposal: what about creating a separate directory, parallel to releases, called e.g. failed-releases/, and moving directories there after a rollback? It could be subject to the same regime of cleaning up more than N entries that the releases directory is subject to. And then you would retain the property that the releases directory only contains releases that got as far as `deploy:restart`.

leehambley:
I was under the impression that v3 didn't delete failed released by default.

jsha:
Good point, thanks. I'll try upgrading to v3.

PR #1902, opened 2017-06-30
Add delete rollback source option

igaiga:
### Summary

I suggest delete source option when using deploy:rollback.

The reason is same as follow.

`When using unicorn and similar servers that do 0 downtime deploys by pre-loading a new set of server app instances - removing the old deployed directory will cause errors until the new ones have booted.`
issue #547 

- set :delete_rollback_source, true  #=> delete
- set :delete_rollback_source, false #=> not delete
- not set(default) #=> delete (same as current)

### Short checklist

### DIFF
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -43,7 +43,7 @@ namespace :deploy do
   end
 
   task :finishing_rollback do
-    invoke "deploy:cleanup_rollback"
+    invoke "deploy:cleanup_rollback" if fetch(:delete_rollback_source, true)
   end
 
   task :finished do


mattbrictson:
@leehambley care to comment on this one? Back in 2013 (!) you wrote on #547:

> I'd also take a patch to make this configurable via a variable.

Personally, given the fact that the maintainers mostly frown upon use of the rollback feature, I would hesitate making any enhancements to it, even in the form of configuration. There is a workaround, since you can [override or clear](http://capistranorb.com/documentation/advanced-features/overriding-capistrano-tasks/) the `deploy:cleanup_rollback` task.

mattbrictson:
BTW, thanks @igaiga for the PR! We'll review and let you know if it is accepted.

mattbrictson:
On further review, I've decided not to accept this PR. There is a simple workaround, which is to add this to your `deploy.rb` if you'd like to disable the deleting of source on rollback:

```ruby
Rake::Task["deploy:cleanup_rollback"].clear_actions
```

PR #1912, opened 2017-08-03
flush  buffer

kminiatures:
### Summary

I did not see ask_question on MinGW, so I added $stdout.flush.


mattbrictson:
Thanks for the PR! This looks good so far. I don't see a need for a test, but could you add a CHANGELOG entry as suggested [above](https://github.com/capistrano/capistrano/pull/1912#issuecomment-319870535)?

kminiatures:
Thank you for your reply ❗️ 
I added changelog so please watch it.

mattbrictson:
Thanks! 🙌

Issue #1884, opened 2017-04-26
Better explain how to specify server-side environment variables


mattbrictson:
Applications deployed with Capistrano often need to have secret keys, database URLs, etc. specified via environment variables. This comes up a lot on Stack Overflow and should be better explained in the official Capistrano docs.

**capistrano + sidekiq / unable to pick environments variables**
http://stackoverflow.com/a/43614787/4625365

**Capistrano and API keys in ENV variables?**
http://stackoverflow.com/a/39628182/4625365

leehambley:
@mattbrictson good point… do we literally mean https://12factor.net/, or more in the direction of "how to get secrets onto a server without putting them in plaintext in the repo" ?

mattbrictson:
I think pointing to 12 Factor is good, because we don't necessarily want to be in the business of completely documenting everything relating to deployment. But additionally I think some concrete examples would help.

cefigueiredo:
I think that the most troublesome is that Capistrano deploys don't use `.bashrc` or `.bash-profile` (not sure about others shells...)
That turns difficult setup things... often being necessary a tool like https://github.com/laserlemon/figaro to setup environment variables on `shared/config` folder

leehambley:
@cefigueiredo it's by design, after years of hard to diagnose issues that we specifically document why not handling shell init files. Capistrano has always walked a fine line between a deployment tool., and a provisioning tool. Handling shell init files makes the process more difficult than it ought to be and causes more harm than good. We find in general, that people who find this to be a problem also know what's going on, and how to work around it such as yourself.

mattbrictson:
On Ubuntu, `.bashrc` works just fine.

http://stackoverflow.com/questions/43606285/capistrano-sidekiq-unable-to-pick-environments-variables/43614787#43614787

mattbrictson:
But I agree this a problem that trips a lot of people up, so it is worth explaining it better in the docs.

leehambley:
If that's the case Matt (thanks for the SO link) we might need to find out why - because it shouldn't as far as I know - http://capistranorb.com/documentation/faq/why-does-something-work-in-my-ssh-session-but-not-in-capistrano/

We specifically document that Cap goes out of it's way to ensure a non-login, non-interactive shell, that should mean that nothing is loaded, unless Ubuntu is setting `$BASH_ENV`?

mattbrictson:
Perhaps this is the reason (copied from the bash manpage):

[...]

Issue #1870, opened 2017-03-24
Problems when remote user has tcsh or csh as default shell


kperry42:
I'm attempting to upgrade from Cap 2.x to Cap 3.x - everything worked fine in 2.x.  With version 3, I'm seeing the following issue:  For legacy reasons, the account I'm using on the remote servers that I'm deploying to has tcsh as its login shell.  I set a couple of envars in :default_env in my config/deploy.rb.  When I run, say, "cap dev deploy", it fails.  

Every command that it tries to execute - including such basics as "deploy:check:directories" produces a (non-fatal, but problematic) "export: Command not found" error.  When it tries to read the contents of the REVISION file in the "deploy:set_previous_revision" task, this happens:
```
 DEBUG [294ec78e] Command: ( export COMPOSER_DISCARD_CHANGES="true" PATH="/opt/rh/rh-php56/root/usr/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/usr/local/apps/nodejs/current/node/bin" ; /usr/bin/env cat /usr/local/apps/timeline/current/REVISION 2>/dev/null )

 DEBUG [294ec78e]       export: Command not found.

 DEBUG [294ec78e]       cat: 2: No such file or directory
```
So there appear to be two major issues: the lack of an "export" function in tcsh, and the different syntax for routing stderr (tcsh doesn't know what to do with the "2>").

As best I can tell from searching the docs and the web, the only way to get around this at present is to change the user's login shell to bash.  This requirement should at least be well-documented; but preferably, there would be a way to tell Capistrano that the remote user is using tcsh, and have it issue valid tcsh commands.

I mentioned this yesterday in a comment on [sshkit issue 354](/capistrano/sshkit/issues/354#issuecomment-288864314)

(Sorry for the lack of a gist showing this behavior - I don't have a publicly accessible server that's set up this way.)

leehambley:
FWIW from a friend who happend to have an example to hand:

```
[root@pike profile.d]# cat go.csh
set path = ( $path /usr/local/go/bin )
[root@pike profile.d]# cat go.sh
export PATH=$PATH:/usr/local/go/bin
```

Non the less, according to the IEEE Standard 1003.1 (POSIX, 2004) `export` is standard, and the syntax for redirection is specifiied in the same document

 - http://pubs.opengroup.org/onlinepubs/007904975/utilities/export.html
[...]

Issue #1763, opened 2016-08-29
Improve explanation of SSHKit DSL with examples and all available methods


leehambley:
_From @mattbrictson on January 19, 2016 23:58_

From https://github.com/capistrano/capistrano/issues/1531 (which I have closed and moved here instead):

> Perhaps a link to [the SSHKit docs] on the [capistranorb.com] site would be appropriate or perhaps incorporating it into the documentation? Since it's very much a part of the DSL. Perhaps just a list of all the DSL methods available would be great. Would also be great to have the line number/file where they are defined. Perhaps a search feature would also be helpful.

_Copied from original issue: capistrano/documentation#159_

leehambley:
Let's not *move* anything, but we could definitely copy from, and link to it.

Issue #1758, opened 2016-08-29
Recommended use of `ask` is incorrect


leehambley:
_From @mattbrictson on May 31, 2016 19:16_

The [documentation](http://capistranorb.com/documentation/getting-started/user-input/#) recommends this pattern:

``` ruby
set :database_password, ask('Enter the database password:', 'default', echo: false)
```

But this is wrong. The first argument to `ask` is supposed to be a variable name, not a prompt string. Technically the code still works (since the outer `set` overrides the implicit `set` within `ask`), but it is very misleading.

_Copied from original issue: capistrano/documentation#192_

InFog:
Hey there. I want to fix this one. But, as stated, the docs are really misleading.

The signature for `ask` is `ask(key, default=nil, options={})` and in the end it works "fine" because I18n's `default_question` is `question_default: "Please enter %{key} (%{default_value}): "`, so, yes, the "key" is there.

What should be the recommended way to `ask` questions with a prompt string? Or should it just be explicit that an outer `set` would to the job in such situations?

mattbrictson:
There is no official way to `ask` a question with a custom prompt string. The technique currently in the docs is a hack, IMHO, and has no test coverage. My suggestion would be to change the docs so that it only explains these patterns:

```ruby
ask(:database_encoding, "UTF-8")
ask(:database_password, "default", echo: false)
```

If the docs also show an example of what the default prompts look like, that would make this feature is easier to understand.

As an aside, if we want to officially support custom prompts, then perhaps that could be a separate PR. It could be another supported option, like:

```ruby
ask(:database_encoding, "UTF-8", prompt: "My custom question prompt: ")
```

mattbrictson:
Fixed in #1924

PR #1921, opened 2017-10-03
Moved the Official Plugins into a separated page.

InFog:
Also added information to expand the list of 3rd party plugins.

### DIFF
--- a/docs/_includes/navigation.html
+++ b/docs/_includes/navigation.html
@@ -51,17 +51,7 @@
   <li class="divider"></li>
 
   <h5>Plugins</h5>
-  <li><a href="https://github.com/capistrano/bundler/">Bundler</a></li>
-  <li><a href="https://github.com/capistrano/composer/">Composer</a></li>
-  <li><a href="https://github.com/capistrano/drupal-deploy/">Drupal</a></li>
-  <li><a href="https://github.com/capistrano/laravel/">Laravel</a></li>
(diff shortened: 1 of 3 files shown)


InFog:
Actually missing some comments here. I think this way shows more clear which plugins are maintained officially.

Another small issue I had: It was not possible to use the dependencies on `Gemfile.lock`, so I had to update them (I guess due to old versions). Should I add this update to the merge request?

mattbrictson:
This is a nice improvement, thanks!

Regarding the `docs/Gemfile.lock`: please submit a patch as a separate PR. Updates to the docs lock file can have unpredictable effects on the rendering of the site, so I'd like to give that a careful review.

InFog:
Cool. Thank you, @mattbrictson!

I created the PR updating `Gemfile.lock` at #1923.

InFog:
Btw, @mattbrictson anything else that bothers people on the docs? Since this is my second commit there I think I can do some more ;-)

mattbrictson:
Hi @InFog, thanks for the offer! If you browse the [documentation](https://github.com/capistrano/capistrano/issues?q=is%3Aissue+is%3Aopen+label%3Adocumentation) label you might find some other chores to work on. The ones I'd particularly like to see solved are #1884, #1870, #1763, and #1758. The last 2 are probably the easiest to fix.

Any help you can provide would be much appreciated! 🙇

InFog:
Thanks, @mattbrictson. I will take a look on those.

PR #1922, opened 2017-10-04
Prevents last good release from being deleted on cleanup

dmnelson:
### Summary

Issue: https://github.com/capistrano/capistrano/issues/1907

If there are many failed deployments the relese folders will
still be left there and eventually if deploy:cleanup is
executed it will delete the release that is referenced by the
"current" symlink, potentially causing downtime.

That change will prevent the "current release" from ever being
removed during cleanup.

### Short checklist

- [X] Did you run `bundle exec rubocop -a` to fix linter issues?
- [X] If relevant, did you create a test?
- [X] Did you confirm that the RSpec tests pass?
- [X] If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry?

### Other Information

As of now it will keep the current release IN ADDITION to the last X releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,8 @@ gem "capistrano", :github => "capistrano/capistrano"
 [master]: https://github.com/capistrano/capistrano/compare/v3.9.1...HEAD
 
 * Your contribution here!
+* [#1922](https://github.com/capistrano/capistrano/pull/1922): Prevents last good release from being deleted during cleanup if there are too many subsequent failed deploys
 * As of this release, version 2.x of Capistrano is officially End of Life. No further releases of 2.x series are planned, and pull requests against 2.x are no longer accepted. The maintainers encourage you to upgrade to 3.x if possible.
-* Prevents last good release from being deleted during cleanup if there are too many subsequent failed deploys
 
 ## [`3.9.1`] (2017-09-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ gem "capistrano", :github => "capistrano/capistrano"
 
 * Your contribution here!
 * As of this release, version 2.x of Capistrano is officially End of Life. No further releases of 2.x series are planned, and pull requests against 2.x are no longer accepted. The maintainers encourage you to upgrade to 3.x if possible.
+* Prevents last good release from being deleted during cleanup if there are too many subsequent failed deploys
 
 ## [`3.9.1`] (2017-09-08)
 
--- a/features/deploy.feature
+++ b/features/deploy.feature
@@ -60,6 +60,16 @@ Feature: Deploy
     Then 3 valid releases are kept
     And the invalid "new" release is ignored
 
+  Scenario: Cleanup after many failed releases doesn't remove last good release
+    Given config stage file has line "set :keep_releases, 2"
+    And I make 2 deployments
+    And an invalid release named "77777777777777"
+    And an invalid release named "88888888888888"
+    And an invalid release named "99999999999999"
+    When I run cap "deploy:cleanup"
+    Then 3 valid releases are kept
+    And the current directory will be a symlink to the release
+
   Scenario: Rolling Back
     Given I make 2 deployments
     When I run cap "deploy:rollback"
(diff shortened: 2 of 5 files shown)


features/step_definitions/assertions.rb
```diff
@@ -66,7 +66,7 @@
 end
 
 Then(/^the current directory will be a symlink to the release$/) do
-  run_vagrant_command(test_symlink_exists(TestApp.current_path))
+  run_vagrant_command(exists?("e", TestApp.current_path))
```
dmnelson: This will make it check that the symlink points to a valid file/dir besides of just existing

leehambley:
First impression is that this looks good, we've historically had problems in this bit of code (and many proposed solutions involving gymnastics of sorting/etc)

I like your solution mapping over things, but as I don't have a real world project to test it, I'm a little bit at a loss.

Is it possible your server has some canonicalised sorting or collation setting that's atypical?

dmnelson:
@leehambley Thanks a lot for the fast feedback.

Our servers don't have anything unusual regarding sorting, the releases are returned in alphabetical order on https://github.com/capistrano/capistrano/blob/v3.9.1/lib/capistrano/tasks/deploy.rake#L151

The reason this issue has affected us is basically we can't keep a large number of past releases because of storage issues, we only keep 2 as of now. So if deployment fails for any reason, let's say migrations fail, for a couple of times in sequence, what will happen is:

```
1 -> current
2 -> failed deploy
3 -> failed deploy
```

and if we need to run `deploy:cleanup` at that point, it will keep only `2` and `3` because they are the latest ones returned by the `ls` command, and the `current` symlink will point to a nonexistent directory (`1` was deleted).

I don't think this affects most people because they probably keep more releases and won't need to run cleanup to release disk space like our case, but I think that's a good safety net to prevent apps from going down if that ever gets to happen.

will-in-wi:
I'm okay with this in principle. @mattbrictson?

will-in-wi:
The feature tests pass for me with this patch, FYI.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ gem "capistrano", :github => "capistrano/capistrano"
 [master]: https://github.com/capistrano/capistrano/compare/v3.9.1...HEAD
 
 * Your contribution here!
+* Prevents last good release from being deleted during cleanup if there are too many subsequent failed deploys
```
mattbrictson: Can you link to the PR?

features/deploy.feature
```diff
@@ -60,6 +60,16 @@ Feature: Deploy
     Then 3 valid releases are kept
     And the invalid "new" release is ignored
 
+  Scenario: Cleanup after many failed releases doesn't remove last good release
```
mattbrictson: I love tests! Much appreciated 💯

lib/capistrano/tasks/deploy.rake
```diff
@@ -155,11 +155,20 @@ namespace :deploy do
...
+        directories = (valid - valid.last(fetch(:keep_releases))).map do |release|
+          releases_path.join(release).to_s
+        end
+        begin
+          current_release = capture(:readlink, current_path).to_s
+          if directories.include?(current_release)
+            warn t(:wont_delete_current_release, host: host.to_s)
+            directories.delete(current_release)
+          end
+        rescue SSHKit::Command::Failed
```
mattbrictson: This worries me a bit. If the SSH execution fails for *any* reason, this `rescue` swallows it and prints a debug message. I assume this is being done to catch a failure for the `capture` on line 162 if it returns a non-zero exit status. Is there a more robust way to check the existence of the file without resorting to a `rescue`?

dmnelson: That's correct. The idea that comes to mind is to do something like `capture("readlink #{path} || echo '' ")` (so it returns a empty string instead of the error), but that seems a bit hackish to me.

At the same time I'm not familiar enough with the project code to fully understand the risks of that rescue there. Is there another exception that would be safer to catch? Any other suggestions to improve this?

mattbrictson: The idiomatic way to do this in Capistrano is using the `test` command. So you could write:

```ruby
if test("[ -d #{current_path} ]")
  current_release = capture(:readlink, current_path).to_s
  # ...
else
  debug t(:no_current_release, host: host.to_s)
end
```

dmnelson: @mattbrictson Awesome. Thanks very much for the suggestion!

mattbrictson:
Looks good! I'll double-check that the feature tests still pass, and then merge it.

PR #1923, opened 2017-10-04
Updated Gemfile.lock for the docs.

InFog:
Updated docs' `Gemfile.lock` to have more up to date dependencies.


will-in-wi:
I'm in favor of this. @mattbrictson, do you see any problems with doing this?

leehambley:
As ever the only issue with the docs is consistent rendering, but I'm all in favour of this too.

PR #1924, opened 2017-10-05
Fixed wrong `ask` usage examples.

InFog:
As discussed in #1758

### DIFF
--- a/docs/documentation/getting-started/user-input/index.markdown
+++ b/docs/documentation/getting-started/user-input/index.markdown
@@ -7,7 +7,7 @@ User input can be required in a task or during configuration:
 
 ```ruby
 # used in a configuration
-set :database_name, ask(:database_name, "default_database_name")
+ask(:database_name, "default_database_name")
 
 # used in a task
 desc "Ask about breakfast"
@@ -24,7 +24,7 @@ input from being displayed. This option should be used to ask the user for
 passwords and other sensitive data during a deploy run.
 
 ```ruby
-set :database_password, ask(:database_password, 'default_password', echo: false)
+ask(:database_password, 'default_password', echo: false)
 ```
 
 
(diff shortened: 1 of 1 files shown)


docs/documentation/getting-started/user-input/index.markdown
```diff
@@ -7,7 +7,7 @@ User input can be required in a task or during configuration:
 
 ```ruby
 # used in a configuration
-set :database_name, ask('Enter the database name:')
+set :database_name, ask(:database_name, "default_database_name")
```
mattbrictson: The correct syntax would be:

```ruby
ask(:database_name, "default_database_name")
```

(no `set`)

docs/documentation/getting-started/user-input/index.markdown
```diff
@@ -24,15 +24,16 @@ input from being displayed. This option should be used to ask the user for
 passwords and other sensitive data during a deploy run.
 
 ```ruby
-set :database_password, ask('Enter the database password:', 'default', echo: false)
+set :database_password, ask(:database_password, 'default_password', echo: false)
```
mattbrictson: Likewise here

mattbrictson:
I would also mention something like this:

> Note that `ask` itself does not prompt the user immediately. The question is deferred until the first time `fetch` is used to obtain the setting. That means you can `ask` for many variables, but only the variables used by your task(s) will actually prompt the user for input.

The way the examples are written right now it is implied that `ask` immediately displays a prompt, which is not the case.

InFog:
Ops, I forgot to remove the redundant `set` usages.

Done now. Also added the info about `ask` waiting for the first `fetch` to actually ask the question. I didn't know that to be honest :-) So, yes, it is good to point that out in the docs.

mattbrictson:
Looks great, thanks! 🎉

InFog:
:tada:

PR #1925, opened 2017-10-05
Removed prism (for syntax highlight).

InFog:
The highlighter "rouge" is already doing a good job and prism is not
working fine on the docs currently, and this causes the not highlighted
code.

It is pretty safe to rely only on Jelyll's highlighting features instead
of adding another layer "only for the highlight".

I also removed from the code the usage of `/js/vendor/custom.modernizr.js`, since this file is not there.


will-in-wi:
This looks good to me. I don't see much if any difference in code highlighting when I compare the two. Thanks!

InFog:
Thanks, @will-in-wi!

PR #1926, opened 2017-10-07
Initial attempt at a cookbook

will-in-wi:
Per my suggestion in https://github.com/capistrano/rails/issues/199#issuecomment-333403472

Also bringing in the case from https://github.com/capistrano/rails/pull/209

Thoughts?

### DIFF
--- a/docs/documentation/tasks/rails/index.markdown
+++ b/docs/documentation/tasks/rails/index.markdown
@@ -10,7 +10,7 @@ Many of these tasks probably require [Capistrano::Rails](https://github.com/capi
 From [Capistrano/Rails PR #209](https://github.com/capistrano/rails/pull/209)
 
 ```ruby
-namespace :ruby do
+namespace :deploy do
   desc 'Runs any rake task, cap deploy:rake task=db:rollback'
   task rake: [:set_rails_env] do
     on release_roles([:db]) do
@@ -26,6 +26,10 @@ end
 
 Passes in the rake task to be run via an environment variable. Also a simple example of running a rake task on the server.
 
+```bash
+bundle exec cap production deploy:rake task=db:seed
+```
+
 
 ### Conditional migrations
 
@@ -34,13 +38,11 @@ Arising from [Capistrano/Rails issue #199](https://github.com/capistrano/rails/i
 A frequent issue on deploy are slow migrations which involve downtime. In this case, you often want to run the migrations conditionally, where the main deploy doesn't run them, but you can do so manually at a better point. To do so, you could put the following in your `Capfile`:
 
 ```ruby
-if ENV['RUN_MIGRATIONS'].present?
-  require 'capistrano/rails/migrations'
-end
+require 'capistrano/rails/migrations' if ENV['RUN_MIGRATIONS']
 ```
 
 Now the migrations do not run by default, but they will run with the following command:
 
 ```bash
-RUN_MIGRATIONS=true bundle exec cap production deploy:migrate
+RUN_MIGRATIONS=1 bundle exec cap production deploy:migrate
 ```
--- a/docs/documentation/tasks/rails/index.markdown
+++ b/docs/documentation/tasks/rails/index.markdown
@@ -1,15 +1,7 @@
 ---
-title: Task Cookbook
+title: Custom Rails Tasks
 layout: default
 ---
-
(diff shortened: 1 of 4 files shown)


InFog:
Looks good, @will-in-wi.

How should we organize the cookbook for other technologies? Maybe a Cookbook page for Rails, another for Symfony, another for Laravel, etc?

will-in-wi:
Good question. My thought was to start with a single page and h2 headers separating the categories. If that gets unwieldy, we can easily split it out.

InFog:
What about making "Tasks" a title on the menu and adding the cookbooks' pages per "category"? Like Rails, etc? It is more maintainable and we don't need changes that would break indexes like Google in the future. Just an idea.

will-in-wi:
I've reworked it per @InFog's feedback. It's now a category of its own.

InFog:
Nice! It looks good. I cannot say much about the Rails page because I don't have the experience, but the structure is good.

What about adding the links also on the Cookbooks' index? Makes it easier for browsing on mobile ;-)

will-in-wi:
Good idea. Added.

docs/documentation/tasks/rails/index.markdown
```diff
@@ -0,0 +1,46 @@
...
+```ruby
+if ENV['RUN_MIGRATIONS'].present?
+  require 'capistrano/rails/migrations'
+end
+```
+
+Now the migrations do not run by default, but they will run with the following command:
+
+```bash
+RUN_MIGRATIONS=true bundle exec cap production deploy:migrate
```
mattbrictson: The value `true` in this example implies that `false` has the opposite effect, which is not the case. `RUN_MIGRATIONS=1` might be better.

docs/documentation/tasks/rails/index.markdown
```diff
@@ -0,0 +1,46 @@
...
+        with rails_env: fetch(:rails_env) do
+          execute :rake, ENV['task']
+        end
+      end
+    end
+  end
+end
+```
+
+Passes in the rake task to be run via an environment variable. Also a simple example of running a rake task on the server.
```
mattbrictson: Might be good to add an example here like you did below in the migrations section. E.g.

```
bundle exec cap production deploy task=db:seed
```

Aside: does Rake automatically turn key=value arguments into `ENV` values? That is a bit unexpected.

docs/documentation/tasks/rails/index.markdown
```diff
@@ -0,0 +1,46 @@
...
+
+Many of these tasks probably require [Capistrano::Rails](https://github.com/capistrano/rails).
+
+### Run arbitrary rake tasks from environment variables
+
+From [Capistrano/Rails PR #209](https://github.com/capistrano/rails/pull/209)
+
+```ruby
+namespace :ruby do
+  desc 'Runs any rake task, cap deploy:rake task=db:rollback'
```
mattbrictson: Should this namespace be `:ruby` or `:deploy`?

I like to encourage people to avoid the `:deploy` namespace because that way you can tell what tasks come out of the box vs those you've added yourself.

docs/documentation/tasks/rails/index.markdown
```diff
@@ -0,0 +1,46 @@
...
+### Conditional migrations
+
+Arising from [Capistrano/Rails issue #199](https://github.com/capistrano/rails/issues/199)
+
+A frequent issue on deploy are slow migrations which involve downtime. In this case, you often want to run the migrations conditionally, where the main deploy doesn't run them, but you can do so manually at a better point. To do so, you could put the following in your `Capfile`:
+
+```ruby
+if ENV['RUN_MIGRATIONS'].present?
+  require 'capistrano/rails/migrations'
+end
```
mattbrictson: I would 1-line this as:

```ruby
require 'capistrano/rails/migrations' if ENV['RUN_MIGRATIONS']
```

Also, `present?` is probably unavailable here because it comes from ActiveSupport, which is not loaded by `cap`.

docs/documentation/tasks/rails/index.markdown
will-in-wi: Will do. Per Rake handling environment variables, it sure does: https://github.com/ruby/rake/blob/master/lib/rake/application.rb#L741

will-in-wi:
I've addressed all feedback thus far. Any further comments before merge?

InFog:
👍

PR #1928, opened 2017-10-09
Replace flatten with flat map

poctek:
As mentioned in [this Rails PR](https://github.com/rails/rails/pull/14240), `#flat_map` is faster than `#map` and then `#flatten`.

### Summary

(Guidelines for creating a bug report are available
here: https://github.com/capistrano/capistrano/blob/master/DEVELOPMENT.md)

Provide a general description of the code changes in your pull
request... were there any bugs you had fixed? If so, mention them. If
these bugs have open GitHub issues, be sure to tag them here as well,
to keep the conversation linked together.

### Short checklist

### Other Information

If there's anything else that's important and relevant to your pull
request, mention that information here.

If you are updating any of the CHANGELOG files or are asked to update the
CHANGELOG files by reviewers, please add the CHANGELOG entry at the top of the file where indicated.

Thanks for helping improve Capistrano!

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -71,7 +71,7 @@ Pull requests are awesome, and if they arrive with decent tests, and conform to 
 
 Your code should conform to these guidelines:
 
- * The code is MIT licenced, your code will fall under the same license if we merge it.
+ * The code is MIT licensed, your code will fall under the same license if we merge it.
  * We can't merge it without a [good commit message](http://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message). If you do this right, Github will use the commit message as the body of your pull request, double win.
  * If you are making an improvement/fix for an existing issue, make sure to mention the issue number (if we have not yet merged it )
  * Add an entry to the `CHANGELOG` under the `### master` section, but please don't mess with the version.
--- a/lib/capistrano/configuration/host_filter.rb
+++ b/lib/capistrano/configuration/host_filter.rb
@@ -3,8 +3,7 @@ module Capistrano
     class HostFilter
       def initialize(values)
         av = Array(values).dup
-        av.map! { |v| v.is_a?(String) && v =~ /^(?<name>[-A-Za-z0-9.]+)(,\g<name>)*$/ ? v.split(",") : v }
-        av.flatten!
+        av = av.flat_map { |v| v.is_a?(String) && v =~ /^(?<name>[-A-Za-z1-9.]+)(,\g<name>)*$/ ? v.split(",") : v }
         @rex = regex_matcher(av)
       end
 
--- a/lib/capistrano/configuration/role_filter.rb
+++ b/lib/capistrano/configuration/role_filter.rb
@@ -3,8 +3,7 @@ module Capistrano
     class RoleFilter
       def initialize(values)
         av = Array(values).dup
-        av.map! { |v| v.is_a?(String) ? v.split(",") : v }
-        av.flatten!
+        av = av.flat_map { |v| v.is_a?(String) ? v.split(",") : v }
         @rex = regex_matcher(av)
       end
 
--- a/lib/capistrano/doctor/servers_doctor.rb
+++ b/lib/capistrano/doctor/servers_doctor.rb
@@ -56,7 +56,7 @@ module Capistrano
         private
 
         def find_whitespace_roles
-          servers.map(&:roles).map(&:to_a).flatten.uniq
+          servers.map(&:roles).flat_map(&:to_a).uniq
                  .select { |role| include_whitespace?(role) }
         end
       end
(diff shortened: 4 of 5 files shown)


leehambley:
Thanks for opening a PR!

> As mentioned in this Rails PR, #flat_map is faster than #map and then #flatten.

Is this _really_ worth optimizing, or is it simply a "oh, could be better" - I'm loathed to believe that Capistrano can spend enough time here in this method that an order of magnitude speed increase (as mentioned in the linked Rails PR) can really change the Capistrano user experience?

will-in-wi:
I agree that for the likely case of <100 items in the array, this isn't likely to make a measurable difference. I do like that the code is simpler...

will-in-wi:
@poctek: What made you want to create the PR? Did you hit a performance bottleneck with it?

Thanks!

will-in-wi:
As a side note, it looks like there is a difference between `#flatten` and `#flat_map`. If this were `#flatten(1)`, they would be the same. It's possible that it still doesn't matter, but it's worth double-checking before merge.

mattbrictson:
My 2¢: I would guess that performance is not impacted in any meaningful way. But if code readability is improved and there is no risk of regression, then I still see some value.

However I would *not* include something like this in the CHANGELOG. It is not a user-facing change.

poctek:
@will-in-wi I've recently started working with capistrano (awesome tool, btw)) and faced some problems with whenever gem integration today. Now I've solved the problem, but noticed that `map` might've been replaced by the `flat_map` while was reading the role filter source code.
Searched for the other occurrences, replaced them and checked if everything was working. If you think that you shouldn't merge the PR, it's OK - it didn't take too much time to make me regret)

leehambley:
Thanks all, I agree it's more readable, if we can address Matt's
observation about flatten(1) Vs flatten() I'd be glad to gratefully accept
the PR. I'll re-read the docs and see if the usage should match and then
comment back.

leehambley:
I stepped through all instances of calls to `HostFilter.initialize` with byebug today, and as far as our tested behaviour goes there's no changes (as one would expect) to the results before/after this patch.

I'm reminded how this code works, and why it works the way it does, it typically receives either one string or an array of strings. Incase it's one string we split it on command and re-join it, hence in the case that we would be called with `["one", "two,three", "four"]` we should correctly expand it to a "flat" array of four items.

With or without this change, it works correctly, though we don't test this "pathological" case, we test for simple cases (sometimes containing complex regexes, but that doesn't concern us at this point in the code).

```
    2: module Capistrano
    3:   class Configuration
    4:     class HostFilter
    5:       def initialize(values)
    6:         debugger
=>  7:         av = Array(values).dup
    8:         av = av.flat_map { |v| v.is_a?(String) && v =~ /^(?<name>[-A-Za-z1-9.]+)(,\g<name>)*$/ ? v.split(",") : v }
    9:         @rex = regex_matcher(av)
   10:       end
   11:
(byebug) av = Array(["one", "two,three", "four"]).dup
["one", "two,three", "four"]
(byebug) av.flat_map { |v| v.is_a?(String) && v =~ /^(?<name>[-A-Za-z1-9.]+)(,\g<name>)*$/ ? v.split(",") : v }
["one", "two", "three", "four"]
```

I'd say this PR can be merged as-is. I'd ask @poctek if he's interested in extending the test suite to include a new test for this case (above).

I'd be happy to see this merged with or without an extra test-case, I defer to anyone else to 👍 👎 me and I'll do whatever we decide.

leehambley:
```
commit 7e6db91daf0ce7ba5544b2304d3520a06530645b
Author: Lee Hambley <leehambley@harrow.io>
Date:   Wed Oct 11 15:13:14 2017 +0200

    Introduce test to check valid, but unusual use of hostfilter

diff --git a/spec/lib/capistrano/configuration/host_filter_spec.rb b/spec/lib/capistrano/configuration/host_filter_spec.rb
index b20d2424..1bf402ad 100644
--- a/spec/lib/capistrano/configuration/host_filter_spec.rb
+++ b/spec/lib/capistrano/configuration/host_filter_spec.rb
@@ -41,6 +41,11 @@ module Capistrano
           it_behaves_like "it filters hosts correctly", %w{server1 server3}
         end
 
+        context "with mixed splittable and unsplittable strings" do
+          let(:values) { %w{server1 server2,server3} }
+          it_behaves_like "it filters hosts correctly", %w{server1 server2 server3}
+        end
+
         context "with a regexp" do
           let(:values) { "server[13]$" }
           it_behaves_like "it filters hosts correctly", %w{server1 server3}

```

Here's the test (passes, with and without the patch), but I can't push on `poctek/capistrano` so it doesn't show up. I pushed a branch with the same name including both his changes and mine.

poctek:
@leehambley updated the branch with the new test. Thank you!

leehambley:
@poctek thanks kindly for applying some pressure here, who knows if it's faster, but I consider having improved the test suite and simplified code to be a worthwhile win.

As @mattbrictson didn't think we needed a changelog entry, I'm fine - merging now.

will-in-wi:
Thanks, all!

Issue #1929, opened 2017-10-09
`#lock` is undocumented`


will-in-wi:
Examples of usage are here: https://github.com/capistrano/capistrano/blob/bf45cbafe5ef396933c81f334bdbc8c6d32a38c2/spec/lib/capistrano/version_validator_spec.rb

We probably want to describe what it possible, and describe various ways of using it, as well as a brief summary of what it is for (handling the case where you don't use bundler, or do use bundler without bundle exec).

will-in-wi:
Responding to @mattbrictson's question from https://github.com/capistrano/capistrano/issues/1056#issuecomment-335307196

> Would it also be nice if the deploy.rb generated by cap install used lock "~> MAJOR.MINOR" by default? Instead of lock VERSION?

I think this would be preferable. I can assemble a PR if others think this makes sense.

will-in-wi:
I'm not sure where docs for this would go… It seems a minor thing which doesn't need its own page…

will-in-wi:
Here's a first draft. Thoughts? https://github.com/capistrano/capistrano/pull/1931

PR #1931, opened 2017-10-10
Initial draft of version locking documentation

will-in-wi:
Documentation per https://github.com/capistrano/capistrano/issues/1929

### DIFF
--- a/docs/_includes/navigation.html
+++ b/docs/_includes/navigation.html
@@ -27,6 +27,7 @@
   <li><a href="/documentation/getting-started/flow/">Flow</a></li>
   <li><a href="/documentation/getting-started/rollbacks/">Rollbacks</a></li>
   <li><a href="/documentation/getting-started/cold-start/">Cold Start</a></li>
+  <li><a href="/documentation/getting-started/version-locking/">Version Locking</a></li>
   <li><a href="/documentation/getting-started/tasks/">Tasks</a></li>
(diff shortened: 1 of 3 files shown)


leehambley:
> Even today, it is easy to run Capistrano without `bundle exec` or a binstub (`bin/cap`, obtained through `bundle binstub capistrano`), resulting in unexpected behavior.

Reads really well barring the american english :trollface:  (just kidding, we standardi*z*ed on US english at some point I think)

If memory serves the version comparator comes from Rubygems - looks like we use https://ruby-doc.org/stdlib-2.4.2/libdoc/rubygems/rdoc/Gem/Dependency.html#method-i-3D-7E here https://github.com/capistrano/capistrano/blob/master/lib/capistrano/version_validator.rb#L16 - maybe we should link the docs, this version comparator works totally differently in Ruby for example from Node.js (which bites me once a week)

will-in-wi:
We could go with mid-Atlantic English and try to have a 50/50 split of U.S. and U.K. English spellings. 😄

Heh, when I looked at the code, I just assumed that strings like `~> 3.0.9` were some magic type of Regex I'd never heard of. I'm kind of surprised that they overrode that operator… Either way, really good to know, I'll update the docs.

will-in-wi:
I've pushed the clarification.

leehambley:
LGTM

PR #1933, opened 2017-10-11
Clarify shell requirement. Fixes #1870

### LINKED ISSUE #1870: Problems when remote user has tcsh or csh as default shell

will-in-wi:
Documentation fix for https://github.com/capistrano/capistrano/issues/1870.

Clarifies that a POSIX shell is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ Likewise, your server(s) will likely need supporting software installed before y
 
 Capistrano is designed to deploy using a single, non-privileged SSH user, using a *non-interactive* SSH session. If your deployment requires `sudo`, interactive prompts, authenticating as one user but running commands as another, you can probably accomplish this with Capistrano, but it may be difficult. Your automated deployments will be much smoother if you can avoid such requirements.
 
+#### Shells
+
+Capistrano 3 expects a POSIX shell like Bash or Sh. Shells like tcsh, csh, and such may work, but probably will not.
+
 ## Quick start
 
 ### Requirements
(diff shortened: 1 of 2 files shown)


leehambley:
LGTM (til that those other shells can be /really/ broken)

PR #1935, opened 2017-10-11
Add commented out suggestion to default deploy.rb template. Fixes #1896

### LINKED ISSUE #1896: Known, non-critical, security vulnerability re: verification of host keys
After discussing this security vulnerability in private, the capistrano team have come ".. to the conclusion that this is not a critical security vulnerability .." and have asked me to open an issue.

**tl;dr workaround**: add `set :ssh_options, paranoid: :secure` to `config/deploy.rb`.

# Original Vulnerability Report

Hi Matt,

As you know, to be secure, Capistrano and the people who use it must configure SSH correctly. Some configuration options are the responsibility of the users, and others are the responsibility of Capistrano.

When using SSH (in any context) the host key of the server must be verified. Failing to do so is a well-known vulnerability leading to DNS-based person-in-the-middle attacks. This attack is so well-known that it is mentioned in the manual page for SSH.

>
> Additionally, the server must be able to verify the client's host key (see the description of /etc/ssh/ssh_known_hosts and ~/.ssh/known_hosts, below) for login to be permitted.  This authentication method closes security holes due to IP spoofing, DNS spoofing, and routing spoofing.
>

Capistrano does not check host keys.

There is a workaround. Users can `set :ssh_options, paranoid: :secure`. This option will be passed, as you know, to net-ssh.

We believe capistrano should set this option by default. Systems should be secure by default. Perhaps there are legitimate use-cases where an insecure configuration is desirable, but security should be the default.

If you agree, please let me know what your timeline will be for releasing a fix.

Thank you for your consideration,
-Jared

# Response from Capistrano Team

Hi Jared,

I’ve discussed this with the other maintainers, and we all came to the conclusion that this is not a critical security vulnerability that warrants requesting a CVE.

Capistrano relies on net-ssh’s default behavior for host key verification, which is trust-on-first-use (and verify thereafter). As you point out, this is not as secure as it could be, but it strikes a good balance. If we were to issue an emergency patch to change this behavior, it would likely break a lot of Capistrano workflows, especially those involving unattended deploys (e.g. via CI).

The maintainers would prefer to move this discussion to a public forum, so that we can weigh these pros and cons and find a solution that works. For example, perhaps we can include a recommended host key verification setting in the deploy.rb that is generated for new projects, with appropriate documentation, while maintaining the current default for legacy projects.

If you are comfortable with continuing this conversation in public, would you mind creating an issue on GitHub?

Once again, thank you for taking the time to surface this issue and for helping to improve Capistrano for everyone!

will-in-wi:
Fixes https://github.com/capistrano/capistrano/issues/1896 by adding a suggested default, commented out normally.

### DIFF
--- a/lib/capistrano/templates/deploy.rb.erb
+++ b/lib/capistrano/templates/deploy.rb.erb
@@ -34,3 +34,6 @@ set :repo_url, "git@example.com:me/my_repo.git"
 
 # Default value for keep_releases is 5
 # set :keep_releases, 5
+
+# Uncomment the following to require manually verifying the host key before first deploy.
+# set :ssh_options, verify_host_key: :secure


will-in-wi:
I don't think this is deserving of a changelog entry. Feel free to disagree.

will-in-wi:
Thanks for the review, @mattbrictson!

PR #1939, opened 2017-10-16
Include :deploy namespace in block example

therubymug:
### Summary

Small tweak to the documentation on before/after hooks.


will-in-wi:
This makes sense to me. Thanks!

PR #1947, opened 2017-10-25
updated remote_file's doc

InFog:
To make it visible it was removed on a previous release.


mattbrictson:
Good catch. Thanks! 🙌

InFog:
👍 🎉

Issue #1058, opened 2014-05-18
If i want git clone in sequence not parallel.


doomzhou:
i can do it with custom task like below ,

    desc "echo2"                                                                                                             
    task :echo2  do                                                                                                          
      on roles(:all), in: :sequence, wait: 2 do |host|                                                                 
        puts "doom #{host}    "                                                                                              
      end                                                                                                                    
     end

but how  **git clone** or **git update** is not clear

rubyrobot:
I have the same problem. Parallelism is a basic feature of SSHKit:

```ruby
on release_roles(:all), in: :groups, limit: 10, wait: 3 do
```
But it seems not to be implemented into Capistrano's default tasks. If you deploy to dozens or hundreds of servers, all updating code from a central git repository at once, the repository server will start dropping ssh connections, making your deploy fail:

```
(Backtrace restricted to imported tasks)
cap aborted!
SSHKit::Runner::ExecuteError: Exception while executing as root@123.123.123.123: git exit status: 128
git stdout: Nothing written
git stderr: ssh_exchange_identification: Connection closed by remote host
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

SSHKit::Command::Failed: git exit status: 128
git stdout: Nothing written
git stderr: ssh_exchange_identification: Connection closed by remote host
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

Tasks: TOP => git:check
(See full trace by running task with --trace)
```

Has anyone solved this? I think there must be a better way than overwriting all of Capistrano's default scm tasks.

leehambley:
You can simply override the one task where we do the clone. IT's a Ruby method, and if you are using Ruby 2.x you can use `Module.prepend`. If you post back with an attempt, I'll help you get it working (and then I/we can document it, as it comes up quite a lot)

rubyrobot:
Thanks for reopening, @leehambley.

I think it's more than the on task where the clone is performed. There is the git:check task, getting current revision etc..

I am sorry, I am not experienced with Rake syntax and overriding behavior. But I know that Rake is additive. Could you please clarify on how to use the prepend? From what I understand the original task should be executed inside a block and the overriding task should be wrapped around it like so:

```ruby
namespace :git do
   
  desc 'Check that the repository is reachable'
  task :check do
    fetch(:branch)
    on release_roles(:all), in: :groups, limit: 10, wait: 3 do
[...]

PR #1957, opened 2017-12-18
(#1058) Update git.rake - configurable max concurrent connections

grzegorzblaszczyk:
Default git_max_concurrent_connections is set to 10 hosts per group.

Default git_wait_interval is set to 3 seconds.

This enables deployments for large number of servers with default Gitlab or similar git server configuration with default SSH settings (max 10 concurrent connections).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ gem "capistrano", :github => "capistrano/capistrano"
 
 [master]: https://github.com/capistrano/capistrano/compare/v3.10.1...HEAD
 
-* Your contribution here!
+### New features:
+
+* [#1058](https://github.com/capistrano/capistrano/pull/1058): Update git plugin rake tasks - configurable max concurrent connections and wait interval settings
 
 ## [`3.10.1`] (2017-12-08)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ gem "capistrano", :github => "capistrano/capistrano"
 
 [master]: https://github.com/capistrano/capistrano/compare/v3.10.1...HEAD
 
+* Your contribution here!
 ### New features:
 
 * [#1058](https://github.com/capistrano/capistrano/pull/1058): Update git plugin rake tasks - configurable max concurrent connections and wait interval settings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ gem "capistrano", :github => "capistrano/capistrano"
 * Your contribution here!
 ### New features:
 
-* [#1058](https://github.com/capistrano/capistrano/pull/1058): Update git plugin rake tasks - configurable max concurrent connections and wait interval settings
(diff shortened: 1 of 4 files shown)


will-in-wi:
https://github.com/capistrano/capistrano/issues/1058

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ gem "capistrano", :github => "capistrano/capistrano"
 [master]: https://github.com/capistrano/capistrano/compare/v3.10.1...HEAD
 
 * Your contribution here!
+### New features:
+
+* [#1058](https://github.com/capistrano/capistrano/pull/1058): Update git plugin rake tasks - configurable max concurrent connections and wait interval settings
```
will-in-wi: The PR ID is #1957. The issue ID is #1058. I think we generally link to the PR ID.

will-in-wi:
Thanks for the contribution!

Heretofore, we've avoided adding a bunch of custom parameters like this, but I'm not sure how the thinking of the other maintainers has changed. I added one comment to the PR, and I'd like some feedback from @mattbrictson and/or @leehambley before merging.

mattbrictson:
As [mentioned in #1058](https://github.com/capistrano/capistrano/issues/1058#issuecomment-335584192), I think that this is a general enough problem that it deserves a general solution.

Although this PR is well written, has good tests, and conforms to the contribution guidelines (thank you!), I don't feel comfortable merging it right now. I would rather see a more general solution, perhaps in SSHKit itself, that allows a user to set a maximum limit on parallelization of the `:parallel` runner.

@grzegorzblaszczyk in the meantime, would something like this work for you as a workaround?

```ruby
# Place this in deploy.rb
SSHKit.config.default_runner_config = { in: :groups, limit: 10, wait: 3 }
```

mattbrictson:
Further discussion: https://github.com/capistrano/sshkit/issues/414

grzegorzblaszczyk:
@mattbrictson 

Thanks for your suggestion but your workaround affects *every* command sent to each server and I wanted to be able to only limit git-related tasks, so only SCM plugin would be limited.

grzegorzblaszczyk:
@mattbrictson Any update on this one? Until now I need to increase the number of concurrent SSH connections on git server (like Gitlab) while deploying to a large number of hosts (10+).

leehambley:
I'm not actually sure why we lost the momentum on this for 18 months except perhaps oversight on our part, so my apologies for that.

We had another patch contributed this week (fixing shell escaping for SVN user credentials) - so we might be due a new release soon.

grzegorzblaszczyk:
@leehambley  Should I rebase with master or do anything to make your life easier to merge this PR into master?

leehambley:
No need, I just haven't found time yet, sorry - but whether I can carve out 5 or 15 min I will take care of it soon! 🤞

CHANGELOG.md
leehambley: Can you please change this @grzegorzblaszczyk Will is right

grzegorzblaszczyk: @leehambley I think I put the correct link now.

grzegorzblaszczyk:
Rebased version of this is in #2044 , so closing this one.

PR #1959, opened 2018-01-04
remove comment about secrets

uuushiro:
### Summary
secrets.yml will be deprecated and instead adopt credentials.yml.enc in Rails5.2.
https://github.com/rails/rails/pull/30067

credentials.yml.enc is encrypted so it can commit to Git repository.

### Short checklist

- [ ] Did you confirm that the RSpec tests pass?

### DIFF
--- a/lib/capistrano/templates/deploy.rb.erb
+++ b/lib/capistrano/templates/deploy.rb.erb
@@ -21,7 +21,7 @@ set :repo_url, "git@example.com:me/my_repo.git"
 # set :pty, true
 
 # Default value for :linked_files is []
-# append :linked_files, "config/database.yml", "config/secrets.yml"
+# append :linked_files, "config/database.yml"
 
 # Default value for linked_dirs is []
 # append :linked_dirs, "log", "tmp/pids", "tmp/cache", "tmp/sockets", "public/system"


leehambley:
Would it not make sense to replace `secrets.yml` with the new file rather than just removing it ?

mattbrictson:
Thanks for the PR! This seems like a reasonable change; we don't want to promote something in the examples that is deprecated.

I do see some value is showing that `append` can take more than 1 value. If we remove the second value in the example, the example is a little less complete.

That said, I can't think of another example value that is as universal as `database.yml`, so maybe this is fine.

Do any other maintainers have concerns before I merge this?

mattbrictson:
> Would it not make sense to replace secrets.yml with the new file rather than just removing it ?

The new file is encrypted and intended to be committed to SCM, so it would not be in `:linked_files`. That is my understanding, although I have not used the feature.

leehambley:
Fair enough, no objections - seems a bit weird, maybe capistrano/rails needs to do something about actually decrypting this file at deploy time, incase it exists?

mattbrictson:
Rails 5.2+ apparently looks for a `RAILS_MASTER_KEY` environment variable, and if present, uses that to decrypt the file at runtime.

mattbrictson:
@uuushiro congrats on your first contribution to Capistrano! 🎉

PR #1960, opened 2018-01-07
Happy new year! 🎊

mattbrictson:
Update copyright year (not legally necessary, but just to keep things fresh).


will-in-wi:
Happy New Year!!!!

leehambley:
Happy new year gents! Thanks for carrying my weight in 2017 I couldn't have continued to be involved without you :)

PR #1975, opened 2018-03-29
Fix intermittent test failure due to any? DSL

### LINKED ISSUE #1974: any? DSL raises "ArgumentError: wrong number of arguments"
```
Failures:

  1) Capistrano::DSL checking for presence variable exists, is not an array 
     Failure/Error:
       def any?(key)
         value = fetch(key)
         if value && value.respond_to?(:any?) && value.method(:any?).arity < 1
           value.any?
         else
           !fetch(key).nil?

     ArgumentError:
       wrong number of arguments (given 0, expected 1)
     # ./lib/capistrano/configuration.rb:47:in `any?'
     # ./lib/capistrano/configuration.rb:50:in `any?'
     # ./spec/integration/dsl_spec.rb:386:in `block (3 levels) in <top (required)>'
     # ./spec/integration/dsl_spec.rb:405:in `block (4 levels) in <top (required)>'
```

This seems to be a test ordering issue.

Using Ruby 2.4.3, I can consistently reproduce using the following seed:

```
$ bundle exec rspec --seed 50238
```

mattbrictson:
Capistrano's `any?` DSL method, as a convenience, will delegate a zero-
arity call to `any?` for the object stored in the variable that is being
interrogated. For example, `any?(:ssh_options)` will first fetch the
value of the variable `:ssh_options`, and then call `any?` on it with
zero arguments.

The bug stems from the fact that we were not checking the arity of
`any?` before calling it. So if the object being interrogated has an
`any?` method that requires an argument, the method invocation will fail
with an `ArgumentError`.

For whatever reason, under certain circumstances related to test order,
the stub object provided by Rspec would sometimes have an `any?` method
on it (inherited from `Capistrano::DSL`) that required an argument.
Hence we would occasionally see test failures.

~~This commit fixes the bug by checking the arity of `any?` before calling
it.~~

There isn't a reliable way to check the arity of `any?` (it is `-1` for
forwarded or native methods), so the fix is to simply rescue the
`ArgumentError` and fallback to non-`any?` behavior.

Fixes #1974

### Short checklist

### DIFF
--- a/lib/capistrano/configuration.rb
+++ b/lib/capistrano/configuration.rb
@@ -47,10 +47,14 @@ module Capistrano
     def any?(key)
       value = fetch(key)
       if value && value.respond_to?(:any?)
-        value.any?
-      else
-        !fetch(key).nil?
+        begin
+          return value.any?
+        rescue ArgumentError # rubocop:disable Lint/HandleExceptions
+          # Gracefully ignore values whose `any?` method doesn't accept 0 args
+        end
       end
+
+      !value.nil?
     end
 
     def is_question?(key)


mattbrictson:
This PR is incomplete, because `Array.new.method(:any?).arity` is `-1`, not `0`, on Ruby 2.5.1.

mattbrictson:
Fixed

lib/capistrano/configuration.rb
```diff
@@ -47,10 +47,14 @@ def remove(key, *values)
     def any?(key)
       value = fetch(key)
       if value && value.respond_to?(:any?)
```
will-in-wi: Would `value.method(:any?).parameters == []` be a better way of handling this instead of the rescue? Lighter-weight and only saves us from the specific case we care about.

Also, is there any way to write a reliable test for both cases?

will-in-wi:
As a general thought, have we tried using `rspec --bisect` with a failing seed? If we grab the specific seed that fails, does it fail consistently?

will-in-wi:
Sorry, I'm jumping in without full context. I see that you have found a failing seed here: https://github.com/capistrano/capistrano/issues/1974. Does `--bisect` help on that seed?

lib/capistrano/configuration.rb
mattbrictson: > `value.method(:any?).parameters == []`

That doesn't work for `Array#any?` in Ruby 2.5.1:

```
>> [].method(:any?).parameters
[
    [0] [
        [0] :rest
    ]
]
```

And it gives a different result for forwarded methods:

```
>> class Foo
>>   extend Forwardable
>>   def_delegators :@array, :any?
>> 
>>   def initialize
>>     @array = []
>>   end
>> end
:initialize
>> Foo.new.method(:any?).parameters
[
    [0] [
        [0] :rest,
        [1] :args
    ],
    [1] [
        [0] :block,
        [1] :block
    ]
]
```

So basically we can't trust `arity` or `parameters`. 😄

mattbrictson: > Also, is there any way to write a reliable test for both cases?

Yes, I can rewrite the test so that it does not use `stub` (which makes it inconsistent based on the mocha version).

PR #1977, opened 2018-04-12
There should not be an instance where a git revision should be appended

mmiller1:
I ran into a situation where overriding the capistrano deployment version was necessary due to external dependencies. In this situation, it is possible that multiple deployments (after a failed/cancelled deployment) can append git revisions, making this file unable to be parsed. I'm aware this is a very small edge case, but cannot imagine a scenario where the append operator should be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ gem "capistrano", github: "capistrano/capistrano", require: false
 [master]: https://github.com/capistrano/capistrano/compare/v3.10.1...HEAD
 
 * Your contribution here!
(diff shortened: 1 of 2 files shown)


will-in-wi:
I can't think of an instance where it should append either. I'm fine merging this.

Could you add a changelog entry like Danger commented?

I am curious and a little concerned regarding under what circumstances it would be possible for two deployments to be deploying into the same releases folder. That would seem to open up all sorts of problems.

mattbrictson:
Looks good. Thanks!

PR #2003, opened 2018-10-08
Update docs to reflect true default for :local_user

phallstrom:
A change was made to the default for :local_user in

https://github.com/capistrano/capistrano/commit/d947b25aa2f788a9cf083ab3d8c2e92675507740

The docs weren't updated to reflect the new reality. They are now.


mattbrictson:
Great catch. Thanks! 🙇

PR #2005, opened 2018-10-12
Remove dead link from the sidebar.

berkes:
Site is gone and the content seems lost. If someone has a mirror, that
could be re-added.

### Summary

Small fix for the documentation. In which a navigation-item that leads to a dead link, is removed.

### Short checklist


will-in-wi:
This makes sense to me.

@leehambley It looks like you've redirected your entire domain to LinkedIn. Is this correct and we should remove this link, or do you have it somewhere else now?

leehambley:
Ohh man, yep, I have a git repo with that content still hosted, do you
think it still makes sense to link it?

will-in-wi:
I'll merge this and remove the dead link. If you decide to put this back online, we can reverse this with the new link at that time.

PR #2011, opened 2019-01-14
Fix ruby 2.0 travis failure

mattbrictson:
Ruby 2.0 was failing on Travis because a new version of the i18n gem was released that is not compatible with Ruby 2.0. Pin to an older version of i18n when building against Ruby 2.0 as a workaround.

Also add latest Ruby versions to Travis matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: ruby
 rvm:
-  - 2.5.1
-  - 2.4.4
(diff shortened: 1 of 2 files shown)


leehambley:
We're all getting 💀 - I remember _looking forward_ to Ruby 2.0x when I was young, and here it is falling out of support :D

PR #2013, opened 2019-02-18
Fix #1979  by passing pattern to `split` method

### LINKED ISSUE #1979: Release cleanup fails locally
I run a deploy to a local staging test env the cleanup process fails with the below error:

````
DEBUG [45e8d81a] Running /usr/bin/env ls -x /Library/WebServer/Documents/faces_staging/releases on 127.0.0.1
 DEBUG [45e8d81a] Command: /usr/bin/env ls -x /Library/WebServer/Documents/faces_staging/releases
 DEBUG [45e8d81a]       20180426125410 20180426125440 20180426142907 20180426143153 20180426143509
20180426143538 20180426143659 20180426143721 20180426143811 20180426143921
20180426144219 20180426175342
 DEBUG [45e8d81a] Finished in 0.022 seconds with exit status 0 (successful).
  WARN Skipping cleanup of invalid releases on 127.0.0.1; unexpected foldername found (should be timestamp)
````

When I run the production instance and push to the actual production server the releases are cleaned up correctly as defined by :keep_release.  I have checked all folder and user permissions and they are all correct. This is all done locally on a Mac OSX 10.11.16.

deploy.rb
```
# config valid for current version and patch releases of Capistrano
lock "~> 3.10.2"

set :application, "faces"

set :repo_url, "git@gitlab.XXXXXXXXXXXX.git"
set :branch, 'master'

set :format, :pretty
set :pty, true
set :keep_releases, 2

append :linked_dirs, "images"
```

staging.rb file
```
set :stage, :staging
role :app, %w{127.0.0.1}

server '127.0.0.1', user: ENV['CAP_USER'], roles: %w{app}

set :deploy_to, "/Library/WebServer/Documents/staging/"

SSHKit.config.command_map[:bash] = "/bin/bash"
SSHKit.config.command_map[:composer] = "/usr/local/php5-5.6.30-20170213-115230/bin/php /usr/local/bin/composer.phar"
  

#Deploy tasks
namespace :deploy do

  desc "Update composer dependencies"
  task :composer_install do
    on roles(:app) do

        within release_path do
            info "Running composer"
            execute 'composer', 'install', '--no-dev', '--optimize-autoloader'
        end
    end
  end
  
  after :updated, "deploy:composer_install"
end
```

LMK if there is anything else I need to supply. I have been searching for some time and not finding much at all on the "unexpected foldername found" error.

Thanks,

pareeohnos:
Fixes issue #1979 where `deploy:cleanup` would be completely skipped if there were too many existing releases. This was due to the `split` method not working correctly when the string being split contained both `\n` and `\t` characters. This change passes a pattern to the `split` method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ gem "capistrano", github: "capistrano/capistrano", require: false
 
 [master]: https://github.com/capistrano/capistrano/compare/v3.11.0...HEAD
 
+* [#1979](https://github.com/capistrano/capistrano/issues/1979): Fixed issue where deploy:cleanup would be skipped when too many existing releases
+
 * Your contribution here!
 
 ## [`3.11.0`] (2018-06-02)
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -148,7 +148,7 @@ namespace :deploy do
   desc "Clean up old releases"
   task :cleanup do
     on release_roles :all do |host|
-      releases = capture(:ls, "-x", releases_path).split
+      releases = capture(:ls, "-x", releases_path).split(" ")
       valid, invalid = releases.partition { |e| /^\d{14}$/ =~ e }
 
       warn t(:skip_cleanup, host: host.to_s) if invalid.any?


mattbrictson:
There is no good explanation for _why_ this particular change fixes the bug. According to Ruby's own documentation for `split`:

> split(pattern=nil, [limit]) → an_array
If pattern is nil, the value of $; is used. If $; is nil (which is the default), str is split on whitespace as if ' ' were specified.

In other words, `split` and `split(" ")` are equivalent.

Since they are equivalent, I don't see much risk in merging this PR, but not understanding this fix makes me uneasy.

mattbrictson:
Can you verify once more that `cap deploy` using this particular branch fixes the bug for you? I want to be certain before merging. Thanks!

pareeohnos:
Your call on whether you want to merge this, but as discussed in #1979 this was actually because of some legacy code monkey patching the `split` method, so on standard ruby (and I've just confirmed) this is actually working just fine

mattbrictson:
I'll close this PR since there is no bug here to fix (the problem was caused by a monkey-patch). Thanks for investigating!

PR #2020, opened 2019-05-28
Remove the static path to ssh in the git ssh wrapper to allow for greater platform flexibility

### LINKED ISSUE #1221: Dynamically set path to ssh executable

will-in-wi:
Rebase of https://github.com/capistrano/capistrano/pull/1492. See that for discussion.

Fixes #1221 

/CC @signe

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ gem "capistrano", github: "capistrano/capistrano", require: false
 
 [master]: https://github.com/capistrano/capistrano/compare/v3.11.0...HEAD
 
+* Remove the static path to ssh in the git ssh wrapper to allow for greater platform flexibility - [@signe](https://github.com/signe)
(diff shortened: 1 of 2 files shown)


will-in-wi:
Closing in favor of https://github.com/capistrano/capistrano/pull/2053 which does the same thing.

PR #2030, opened 2019-08-29
Exclude danger to fix Ruby 2.0.0 CI error

### LINKED ISSUE #2029: CI build fails for Ruby 2.0.0
https://travis-ci.org/capistrano/capistrano/jobs/578395369#L314

```
Installing faraday-http-cache 2.0.0
Gem::RuntimeRequirementNotMetError: faraday-http-cache requires Ruby version >=
2.1.0. The current ruby version is 2.0.0.
An error occurred while installing faraday-http-cache (2.0.0), and
Bundler cannot continue.
Make sure that `gem install faraday-http-cache -v '2.0.0' --source
'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  danger was resolved to 6.0.9, which depends on
    faraday-http-cache
The command "bundle install --jobs=1" failed and exited with 5 during .
```

mattbrictson:
`danger` no longer installs cleanly on Ruby 2.0.0. Our CI only runs danger on Ruby 2.5, so there is no reason to install it for older versions.

Fixes #2029

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,3 +30,8 @@ end
 if Gem::Requirement.new("< 2.1").satisfied_by?(Gem::Version.new(RUBY_VERSION))
   gem "i18n", "< 1.3.0"
 end
+
+# We only run danger once on a new-ish ruby; no need to install it otherwise
+if Gem::Requirement.new("> 2.4").satisfied_by?(Gem::Version.new(RUBY_VERSION))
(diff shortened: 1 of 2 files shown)


leehambley:
good call, nice one Matt

Lee Hambley
http://lee.hambley.name/
+49 (0) 170 298 5667

PR #2033, opened 2019-09-04
Use a stub for $stdin during testing

### LINKED ISSUE #2032: Some tests fail if stdin is redirected
---

### Steps to reproduce
Run the tests with `> /dev/null`

### Expected behavior
All the tests passes

### Actual behavior
https://salsa.debian.org/ruby-team/capistrano/-/jobs/304495

---

This is a problem that we've found on Debian a little but more than 1 year ago, I didn't report it earlier because it doesn't look serious.

Some of the tests fail under our testing environment, probably because of input redirection, I started a thread on the ruby team to talk about it and this was the reply I've got from one of the team members:

```
I tried locally: with a simple `gbp buildpackage` the tests pass
consistently. Under sbuild, however, they fail consistently.

Looking at the failure messages, it looks like the failing tests are
somehow poking at stdin. one difference between a simple `gbp
buildpackage` and sbuild is that on the first one stdin is left alone,
while under sbuild stdin is redirected.

tried `gbp buildpackage < /dev/null` and it fails consistently

the problem seems to be in lib/capistrano/configuration/question.rb,
method `gets`. if I remove the `return unless $stdin.tty?` from there,
the tests that failed before with `< /dev/null` now pass, but then there
is a test for that specific behaviour that now fails. I think this is
something to discuss with upstream. the key to reproduce is just to run
the tests with `> /dev/null`
``` 

You can see the log of the failures here:
https://salsa.debian.org/ruby-team/capistrano/-/jobs/304495

And the patch we apply on Debian to circumvent that (disable the failing tests):
https://salsa.debian.org/ruby-team/capistrano/blob/fcd89e1042db23548771112abb8b5daabf32fc5c/debian/patches/skip_tests.patch

Thanks for making capistrano!

mattbrictson:
### Summary

This allows the tests to run even in a situation where stdin isn't available, for example when the test is run with `< /dev/null` as is the case for some packaging scripts.

Fixes #2032

### Short checklist

- [ ] ~If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry?~ (this is not a user-facing fix)

### DIFF
--- a/lib/capistrano/configuration/question.rb
+++ b/lib/capistrano/configuration/question.rb
@@ -36,12 +36,12 @@ module Capistrano
       end
 
       def gets
-        return unless $stdin.tty?
+        return unless stdin.tty?
 
         if echo?
-          $stdin.gets
+          stdin.gets
         else
-          $stdin.noecho(&:gets).tap { $stdout.print "\n" }
+          stdin.noecho(&:gets).tap { $stdout.print "\n" }
         end
       rescue Errno::EIO
         # when stdio gets closed
@@ -58,6 +58,10 @@ module Capistrano
 
       def echo?
         (options || {}).fetch(:echo, true)
+      end
+
+      def stdin
+        (options || {}).fetch(:stdin, $stdin)
       end
     end
   end
--- a/spec/integration/dsl_spec.rb
+++ b/spec/integration/dsl_spec.rb
@@ -356,14 +356,16 @@ describe Capistrano::DSL do
   end
 
   describe "asking for a variable" do
+    let(:stdin) { stub(tty?: true) }
+
     before do
-      dsl.ask(:scm, :svn)
+      dsl.ask(:scm, :svn, stdin: stdin)
       $stdout.stubs(:print)
     end
 
     context "variable is provided" do
       before do
-        $stdin.expects(:gets).returns("git")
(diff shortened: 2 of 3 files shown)


mattbrictson:
This fix is incomplete. There are still 2 failing tests when I run the tests with `< /dev/null`:

```
  1) Capistrano::DSL asking for a variable variable is provided sets the input as the variable
     Failure/Error: expect(dsl.fetch(:scm)).to eq "git"

       expected: "git"
            got: :svn

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -"git"
       +:svn
     # ./spec/integration/dsl_spec.rb:370:in `block (4 levels) in <top (required)>'

  2) Capistrano::DSL asking for a variable variable is not provided sets the variable as the default
     Failure/Error: $stdin.expects(:gets).returns("")

     Mocha::ExpectationError:
       not all expectations were satisfied
       unsatisfied expectations:
       - expected exactly once, not yet invoked: #<IO:0x108a28e20>.gets(any_parameters)
       satisfied expectations:
       - allowed any number of times, invoked once: #<IO:0x108a28d80>.print(any_parameters)
     # ./spec/integration/dsl_spec.rb:376:in `block (4 levels) in <top (required)>'
```

mattbrictson:
All fixed now

PR #2038, opened 2019-10-02
Don't use relative paths on lib load

samueloph:
### Summary

Debian-specific patch to fix some failures with testing in our infrastructure.

I understand this might not be desirable upstream, but decided to PR anyway
so you at least know that we are doing it.

The failure happens because the lib directory is removed during testing and
relative paths were used:
https://wiki.debian.org/Teams/Ruby/Packaging/Tests#Case_eight:_autopkgtest_failure

### Short checklist

- [ ] Did you run `bundle exec rubocop -a` to fix linter issues?
Not-applicable
Everthing is passing
Not needed

Feel free to reject the PR in case it's not applicable.

### DIFF
--- a/spec/lib/capistrano/doctor/environment_doctor_spec.rb
+++ b/spec/lib/capistrano/doctor/environment_doctor_spec.rb
@@ -21,8 +21,8 @@ module Capistrano
 
       describe "Rake" do
         before do
-          load File.expand_path("../../../../../lib/capistrano/doctor.rb",
-                                __FILE__)
+          $:.unshift File.expand_path('../../../../../lib', __FILE__)
+          load 'capistrano/doctor.rb'
         end
 
         after do
--- a/spec/lib/capistrano/doctor/gems_doctor_spec.rb
+++ b/spec/lib/capistrano/doctor/gems_doctor_spec.rb
@@ -45,8 +45,8 @@ module Capistrano
 
       describe "Rake" do
         before do
-          load File.expand_path("../../../../../lib/capistrano/doctor.rb",
-                                __FILE__)
+          $:.unshift File.expand_path('../../../../../lib', __FILE__)
+          load 'capistrano/doctor.rb'
         end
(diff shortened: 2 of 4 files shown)


leehambley:
Rubocop in our CI didn't like these changes, but it was mostly "cosmetic".

Prefer use of `$LOAD_PATH` over `$:` and prefer double quotes unless we rely on single quotes to avoid backslack escapes.

Both of those, half a dozen times each for each of the load lines here.

@mattbrictson @will-in-wi any opinion on merging this or not? I honestly can't remember why it was done exactly this way five years ago when we laid the foundations for this, and would be willing to merge if it doesn't set anyone back, and it helps make some stakeholder's life easier, even if they do have unusual requirements? :man_shrugging:

leehambley:
I think the point is that Debian's build tooling does something weird and
this is the workaround, under "sane" circumstances they should be equivalent

spec/lib/capistrano/doctor/environment_doctor_spec.rb
```diff
@@ -21,8 +21,8 @@ module Doctor
 
       describe "Rake" do
         before do
-          load File.expand_path("../../../../../lib/capistrano/doctor.rb",
-                                __FILE__)
+          $:.unshift File.expand_path('../../../../../lib', __FILE__)
```
mattbrictson: Do we need this line at all? I can't test in the Debian packager but at least for me it is not necessary (the specs still pass without it). Because we are already adding ../lib to the load path here:

https://github.com/capistrano/capistrano/blob/1bfbe82007380594358771021ba7a0fa4c2deff9/spec/spec_helper.rb#L1

samueloph:
hmm, I'm feeling that this is indeed too much specific to the Debian tooling needs, it's probably better to keep it out of upstream, I have no problem in applying the patch on the Debian side, thanks a lot for the consideration.

PR #2040, opened 2019-10-26
Fix %x typo in local tasks example

tiendo1011:
%x has 2 variations: 

- `%x(command)`
- `%x["command"]`

The one specified in the doc is none of the above, so I assume it's a mistake.


mattbrictson:
Nice catch, thanks!

PR #2044, opened 2020-02-03
Limit the number of concurrent git connections to 10, and make it configurable

grzegorzblaszczyk:
Default git_max_concurrent_connections is set to 10 hosts per group.

Default git_wait_interval is set to 3 seconds.

This enables deployments for large number of servers with default Gitlab or similar git server configuration with default SSH settings (max 10 concurrent connections).

### DIFF
--- a/docs/documentation/getting-started/configuration/index.markdown
+++ b/docs/documentation/getting-started/configuration/index.markdown
@@ -150,6 +150,13 @@ The following variables are settable:
   * **default:** `current`
   * Name for `current` link pointing to the newest successful deployment's release folder.
 
+* `:git_max_concurrent_connections`
+  * **default:** `10`
+  * Number of concurrent connections to Git repository - useful when your Git server limits the number of simultaneous connections while using SSH (like Gitlab CE).
+
+* `:git_wait_interval`
+  * **default:** `3`
+  * Number of seconds to wait after you reach the limit of concurrent connections to Git repository server and disconnect afterwards to initialize new connections. This prevents from being cut out of SSH server when you use `fail2ban` or similar software for limiting connections to server.
 
 Capistrano plugins can provide their own configuration variables. Please refer
 to the plugin documentation for the specifics. Plugins are allowed to add or
--- a/lib/capistrano/scm/git.rb
+++ b/lib/capistrano/scm/git.rb
@@ -18,6 +18,8 @@ class Capistrano::SCM::Git < Capistrano::SCM::Plugin
         git_ssh: fetch(:git_wrapper_path)
       }
     }
+    set_if_empty :git_max_concurrent_connections, 10
+    set_if_empty :git_wait_interval, 3
   end
 
   def register_hooks
--- a/lib/capistrano/scm/tasks/git.rake
+++ b/lib/capistrano/scm/tasks/git.rake
@@ -4,7 +4,7 @@ git_plugin = self
 namespace :git do
   desc "Upload the git wrapper script, this script guarantees that we can script git without getting an interactive prompt"
   task :wrapper do
-    on release_roles :all do
+    on release_roles(:all), in: :groups, limit: fetch(:git_max_concurrent_connections), wait: fetch(:git_wait_interval) do
       execute :mkdir, "-p", File.dirname(fetch(:git_wrapper_path)).shellescape
       upload! StringIO.new("#!/bin/sh -e\nexec /usr/bin/ssh -o PasswordAuthentication=no -o StrictHostKeyChecking=no \"$@\"\n"), fetch(:git_wrapper_path)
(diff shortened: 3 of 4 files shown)


leehambley:
Hi @grzegorzblaszczyk thanks for the contribution

> with default SSH settings (max 10 concurrent connections).

On which platform is this the defaults and could you link us to the docs please? I selected the values we have very carefully a number of years ago (like 5+) because of some platform having unusually low limits, which caused really confusing error messages.

I'd like to think the world has advanced somewhat in that time, and that we can indeed raise the defaults!

grzegorzblaszczyk:
Hi @leehambley 
Try Gitlab self-hosted with default settings. SSH settings by default are for 10 concurrent connections.

That's why the workaround is to raise the limit in SSH settings but sometimes it's hard to convince Gitlab admin to change these settings. That is why I created the pull request.

leehambley:
So this patch is to lower the amount of default concurrency in Capistrano to make it compatible with self-hosted GitLab environments, did I get it right?

I assumed you were talking about a platform's default outgoing SSH limit, which is why I queried it.

grzegorzblaszczyk:
Yes, it's about making it compatible with self-hosted GitLab environments (server side in `/etc/sshd_config`).

leehambley:
@mattbrictson @will-in-wi can I get a second opinion on this? I like the limits on the `git.rake` file, I'm not sure about the default values.

Can you think of any problems that might arise?

mattbrictson:
@grzegorzblaszczyk could you rebase this PR against the latest master? This repo doesn't use `CHANGELOG.md` anymore so we need to make sure that part of the PR diff is removed. Thanks!

@leehambley we've had requests over the years to switch the git tasks to `in: :groups` to address this problem. This change seems good to me.

The only potential issue is that existing users upgrading may see a slowdown in their deployments if they have a large fleet of servers and are used to the existing parallelism of the git tasks. As long as we document in the release notes that they will have to increase `:git_max_concurrent_connections`, then there is a easy upgrade path.

I'd have to refresh my memory: do you think that `in: :groups` with a sufficiently high `:limit` is pretty much indistinguishable from `in: :parallel`?

grzegorzblaszczyk:
@mattbrictson Thank you for spotting it. I have rebased it against latest master. `CHANGELOG.md` is now from `master` branch.

mattbrictson:
@grzegorzblaszczyk Thanks for the rebase! One more request: in this PR, could you add an explanation of the new `:git_max_concurrent_connections` and `:git_wait_interval` settings to the docs? Our documentation is not the greatest but every little bit helps. 🙂 

https://github.com/capistrano/capistrano/blob/master/docs/documentation/getting-started/configuration/index.markdown

@leehambley @will-in-wi I'm inclined to merge this PR once the docs are updated. Any objections?

leehambley:
No objections @mattbrictson this looks good.

I'm still having minor reservations about changing the world-wide defaults because one platform happens to have restrictions, but we've made similar calls in the past, and objectively the code is better now that it has bounded parallelism.

Thanks @grzegorzblaszczyk

grzegorzblaszczyk:
@leehambley I have updated the docs.

Feel free to comment if anything is unclear with added description of these properties.

If you are happy with the documentation, I would love to see it in `master` branch ;)

mattbrictson:
Merged! Thanks for the PR 🙌

grzegorzblaszczyk:
@mattbrictson Thank you! Looking forward for the next release of `capistrano` then. 💪

mattbrictson:
Expect a 3.12.0 release by this time next week.

mattbrictson:
3.12.0 is released: https://github.com/capistrano/capistrano/releases/tag/v3.12.0

samueloph:
I'd like to suggest adding support for `:git_max_concurrent_connections` to be set to something that would be equivalent to the old behavior (everything in parallel at the same time), if that would be simple enough to implement.

mattbrictson:
This should give you the same behavior as before:

```ruby
set :git_max_concurrent_connections, 1_000_000 # some sufficiently large number
set :git_wait_interval, 0
```

Fjan:
I just upgraded to Capistrano 3.12.0 and noticed deploys (from Github) were taking twice as long. Using `set :git_wait_interval, 0` indeed fixed it. Why is this no longer the default? 

I've used Capistrano for the better part of a decade without any issues with the old defaults, do we really need that waiting interval? (Perhaps we can at least make it zero when deploying to a single server?)

mattbrictson:
I agree, I didn't anticipate that side effect when this PR was merged. I suspect the vast majority of Capistrano users would be better served by a default wait interval of zero. Those that need a non-zero value can set it.

@Fjan would you like to open a quick PR for that?

Fjan:
Sure, added it here: https://github.com/capistrano/capistrano/issues/2049

PR #2052, opened 2020-03-04
Improve misleading example for ssh_options

rlisowski:
Years ago when I was switching from Capistrano 2 to version 3 I introduced ssh_options #557. 
Unfortunately, as an example path, I pasted my own home path and from then on my Github handle pollutes 37K+ lines of code in random projects. 

rlisowski

I would like to stop it.


mattbrictson:
Looks good! Thanks for the PR.

PR #2053, opened 2020-03-25
Use env to find ssh path

JoeDupuis:
### Summary
Changed the call to /usr/bin/ssh in to /usr/bin/env ssh to make capistrano nixos compatible.
Nixos does not keep the binaries in /usr/bin.

### Short checklist

- [ ] If relevant, did you create a test?  **not applicable**
- [x] Did you confirm that the RSpec tests pass? - **Didn't run the  `bundle exec rake features` ones, I got some problem with virtualbox on my system at the moment that would require rebuilding the kernel**
- [ ] If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry?  -- **is the changelog still used? It's empty**

### DIFF
--- a/lib/capistrano/scm/tasks/git.rake
+++ b/lib/capistrano/scm/tasks/git.rake
@@ -6,7 +6,7 @@ namespace :git do
   task :wrapper do
     on release_roles(:all), in: :groups, limit: fetch(:git_max_concurrent_connections), wait: fetch(:git_wait_interval) do
       execute :mkdir, "-p", File.dirname(fetch(:git_wrapper_path)).shellescape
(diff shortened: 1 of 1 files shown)


leehambley:
LGTM @JoeDupuis, thanks :tada: 

@mattbrictson any objections? This seems like a no-brainer

leehambley:
I just did some digging to check how ubiquitous `env` really is, and learned a little bit.

Making a long story short, it's everywhere part of POSIX.1:

- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html

I had never made the link between `env` (to print print env vars), and `/usr/bin/env` to find a utility on the `PATH` were the same, but seems like they are. :tada:

will-in-wi:
Matt approved of the concept in a PR I threw out that does the same thing.

mattbrictson:
Thanks everyone!

> If you are fixing a bug or introducing a new feature, did you add a CHANGELOG entry? **-- is the changelog still used? It's empty**

Good point. I have opened a PR to fix this: #2054

Issue #2046, opened 2020-02-11
Capistrano prints Ruby 2.7.0 keyword parameters warnings


mattbrictson:
Running the Capistrano feature test suite under Ruby 2.7.0 prints the following warning:

```
lib/capistrano/dsl.rb:36: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
ruby/gems/2.7.0/gems/i18n-1.8.2/lib/i18n.rb:195: warning: The called method `t' is defined here
```

Eric-Guo:
Fixed at #2060

PR #2060, opened 2020-05-08
#2046 Fix Ruby 2.7 keyword parameters warnings and add Ruby 2.7 to CI

Eric-Guo:


leehambley:
Thanks @Eric-Guo :)

Eric-Guo:
Not sure if need I remove 2.3 and below version, largely, [2.4 also end of support](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/) this year.

mattbrictson:
Nice work! Thanks for the contribution, @Eric-Guo 🙌

PR #2063, opened 2020-06-14
Fix Bundler deprecation warning in `rake features`

mattbrictson:
This fixes the following warning when running `rake features` using the latest version of Bundler (2.1.4):

```
[DEPRECATED] `Bundler.with_clean_env` has been deprecated in favor of `Bundler.with_unbundled_env`.
If you instead want the environment before bundler was originally loaded, use `Bundler.with_original_env`
(called at spec/support/test_app.rb:194)
```


mattbrictson:
Fastest PR approval ever 🙌

Issue #2074, opened 2021-01-14
Feature Request: git verify-commit


mohamedhafez:
For those of us that GPG sign our git commits, it would be great to have Capistrano make sure that when we deploy, the commit we are deploying has been signed by a trusted person. (e.g., to prevent against deploying code [in a case like this](https://github.blog/2012-03-04-public-key-security-vulnerability-and-mitigation/))

I've just added the following line at the end of `Capistrano::SCM::Git#update_mirror` to do this:
```ruby
git(:"verify-commit", fetch_revision) if fetch(:verify_commit, false)
```

Let me know what you guys think!

mohamedhafez:
the line above will accept any GPG signature for which you have the public key in your keyring, whether it's marked as trusted or not. I don't really know how to make it do that though aside from capturing the output and checking if it says "WARNING" or not, perhaps someone out there has a suggestion?

leehambley:
Hi @mohamedhafez this is absolutely something you should be able to add by patching the rask tasks.

I'm not sure how familiar with Rake (Capistrano) you are, but you should be able to prepend a check, and define your own variables, and tie this in without requiring a patch in Capistrano itself.

I think we'd definitely take a patch for something like this, I can't quite gauge now whether it's more sensible to shell-out to GPG tooling, or find some Gem that binds, or reimplements PGP for checking the signatures. With us shelling out to Git anyway, maybe it makes sense to shell out to Git's signature verification tools.

If you check the flow of tasks here https://capistranorb.com/documentation/getting-started/flow/ then you can see where you might want to integrate this check by reopening the `task` and plugging in the checks you are seeing.

Please let me know if you need/want review on any WIP PRs, this should be fairly easy to add.

mohamedhafez:
I was thinking that if we are using Git's signature verification tools, then it would make sense to have this be a part of `Capistrano::SCM::Git`, so we inherit all the defaults and the wrapper and environment variables, etc, a user might have set for Git. Although it would be easy to make the verification a separate task that hooked in after `git:update` if you guys think that would be cleaner for a PR.

leehambley:
If you reopen `git:update` by just _defining again_ then you will be able to modify that task and we can definitely take a patch in the Git SCM plugin if/when you think it's ready to go.

mohamedhafez:
Ah cool, didn't realize it worked that way! yeah i am a newbie to Capistrano, and just know the basics of Rake in general. 

[...]

PR #2076, opened 2021-01-19
Add a :verify_commit option

mohamedhafez:
When :verify_commit is set to true, after updating the mirror
`git verify-commit <revision>` will be run on the revision that is
about to be deployed to check if it has a valid signature. This
will allow a user to be certain that the code they are about to
deploy was signed by an authorized author, and that they are not
in danger of deploying malicious code in an attack scenario such as
https://github.blog/2012-03-04-public-key-security-vulnerability-and-mitigation/

discussed in https://github.com/capistrano/capistrano/issues/2074

### Short checklist

- [✔] Did you run `bundle exec rubocop -a` to fix linter issues?
- [✔] If relevant, did you create a test?
- [✔] Did you confirm that the RSpec tests pass?

### DIFF
--- a/docs/documentation/getting-started/configuration/index.markdown
+++ b/docs/documentation/getting-started/configuration/index.markdown
@@ -158,10 +158,9 @@ The following variables are settable:
   * **default:** `0`
   * Number of seconds to wait after you reach the limit of concurrent connections to Git repository server and disconnect afterwards to initialize new connections. This prevents from being cut out of SSH server when you use `fail2ban` or similar software for limiting connections to server.
 
-* `:verify_commit`
+* `:git_verify_commit`
   * **default:** `false`
-  * Whether to check if a valid signature exists on the commit to be deployed.
-  * Currently only implemented for Git.
+  * Whether to check if a valid signature exists on the Git commit to be deployed.
 
 Capistrano plugins can provide their own configuration variables. Please refer
 to the plugin documentation for the specifics. Plugins are allowed to add or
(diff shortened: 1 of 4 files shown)


leehambley:
Thanks for the contribution @mohamedhafez I need to make a bit of time to just check naming consistency, and maybe just re-read our docs and make sure everything all continues to make sense but first impression looks really good.

I also noticed you didn't include an integration test, just units, which I think is fine, but I need to just invest a few more :brain: cycles to make sure I'm comfortable with that.

I'll get back to you really soon, thanks again!

leehambley:
@mattbrictson I just realized I left @mohamedhafez hanging here. Did you have a strong opinion about merging this, or not?

mattbrictson:
> @mattbrictson I just realized I left @mohamedhafez hanging here. Did you have a strong opinion about merging this, or not?

This a really nice PR with docs and tests, so from that perspective it looks good. However I know nothing about the `git verify-commit` command so I can't really speak to the usefulness of this feature or how we will maintain it. I'll leave it up to you. 🙂

docs/documentation/getting-started/configuration/index.markdown
```diff
@@ -158,6 +158,11 @@ The following variables are settable:
   * **default:** `0`
   * Number of seconds to wait after you reach the limit of concurrent connections to Git repository server and disconnect afterwards to initialize new connections. This prevents from being cut out of SSH server when you use `fail2ban` or similar software for limiting connections to server.
 
+* `:verify_commit`
```
leehambley, line 161: Hi @mohamedhafez so I'm more than happy to merge this if we change `:verify_commit` to `:git_verify_commit` in accordance with the existing `git` specific variables documented in this file (also at https://capistranorb.com/documentation/getting-started/configuration/)

That was the only issue I noticed, just consistency. I'm not sure other version control systems have a general purpose commit verification system, at least I searched a bit and didn't find anything, so I think we just focus on Git, for now?

leehambley (changes requested):

docs/documentation/getting-started/configuration/index.markdown
mohamedhafez: @leehambley done! made the changes and its passing all tests

LevonAr:
It looks like it went through. Nice one Mo!

PR #2084, opened 2021-04-29
Add new default webpacker path to linked_dirs

razum2um:
Rails default webpacker directory which contains a file with packs checksum digest will change it's location in order to be not affected by `Rails.cache.clear` command with filestore

After this PR https://github.com/rails/webpacker/pull/3002 will be merged, we'll need to update defaults here as well.

### Short checklist

- [x] run `bundle exec rubocop -a`
- [x] If relevant, did you create a test (not relevant, imo)
- [x] confirm that the RSpec tests pass?

### DIFF
--- a/lib/capistrano/templates/deploy.rb.erb
+++ b/lib/capistrano/templates/deploy.rb.erb
@@ -24,7 +24,7 @@ set :repo_url, "git@example.com:me/my_repo.git"
 # append :linked_files, "config/database.yml"
 
 # Default value for linked_dirs is []
-# append :linked_dirs, "log", "tmp/pids", "tmp/cache", "tmp/sockets", "public/system"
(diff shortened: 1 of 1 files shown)


leehambley:
How does this behave in older Rails projects that aren't using that (or, any) version of webpacker yet?

I notice this only updates the deploy `erb` template, so probably this is harmless for existing projects, but I'd love to hear back from you!

Thanks for your submission!

razum2um:
@leehambley webpacker has merged this; and yeah, it affects only new installation :)

leehambley:
Please add your entry in the Changelog and we'll be glad to merge this :pray:

mattbrictson:
Thanks for your contribution! Sorry for the wait. Our release notes are now auto-generated so there is no need to add a changelog entry.

PR #2097, opened 2021-12-28
Add Ruby 3.1 to CI matrix

mattbrictson:
This PR adds Ruby 3.1 to the Travis CI testing matrix.

Also, since Rubocop currently fails to run on Ruby 3.1, limit Rubocop to running on a single Ruby version (2.7). Running Rubocop on multiple Ruby versions is overkill, anyway.

Finally, change the version we use for running Danger to 2.7 (instead of 2.5), in anticipation of older Ruby versions getting dropped eventually.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: ruby
 rvm:
+  - 3.1
   - 3.0
   - 2.7
   - 2.6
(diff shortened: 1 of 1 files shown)


mattbrictson:
Hmm. Travis-CI.org has apparently been shut down and builds are no longer being run. No point in updating the Travis config, then. 😕 

Closing.

PR #2100, opened 2022-01-08
Switch from Travis to Circle CI

mattbrictson:
Travis-CI.org has been shut down. Replace our Travis config with CircleCI instead. Circle offers a generous amount of build credits for open source repos.


.circleci/config.yml
```diff
@@ -0,0 +1,125 @@
...
+
+executors:
+  ruby:
+    parameters:
+      version:
+        description: "Ruby version number"
+        default: "3.1.0"
+        type: string
+    docker:
+      - image: cimg/ruby:<< parameters.version >>
```
mattbrictson, line 11: This is the preferred docker image to use for Ruby builds. It only has tags for recent Ruby versions, though, so we have to use a "legacy" image for testing older versions of Ruby.

.circleci/config.yml
```diff
@@ -0,0 +1,125 @@
...
+  ruby_legacy:
+    parameters:
+      version:
+        description: "Ruby version number"
+        default: "2.4.10"
+        type: string
+    docker:
+      - image: circleci/ruby:<< parameters.version >>
```
mattbrictson, line 19: Here is the "legacy" docker image.

.circleci/config.yml
```diff
@@ -0,0 +1,137 @@
...
+    steps:
+      - checkout
+      - bundle_install:
+          version: << parameters.version >>
+          bundler: "< 2"
+      - run: bundle exec rake spec
+
+  spec_2_0:
+    docker:
+      - image: ruby:2.0.0-p648
```
mattbrictson, line 96: Circle doesn't have any docker image for Ruby 2.0. So to test 2.0 we have to break the pattern, bypass the Circle convenience images, and use the `ruby` image directly.

Rakefile
```diff
@@ -1,16 +1,19 @@
 require "bundler/gem_tasks"
 require "cucumber/rake/task"
 require "rspec/core/rake_task"
-require "rubocop/rake_task"
```
mattbrictson, line 4: We don't install rubocop on all versions of Ruby, because some versions don't work. So we can't assume rubocop will be present. I have wrapped this in a begin/end to handle that case.

mattbrictson:
@will-in-wi @leehambley any objections to merging this one?

leehambley:
Looks good @mattbrictson thanks for doing the leg work!

PR #2103, opened 2022-02-07
Add `linked_dirs` and `linked_files` to VariblesDoctor WHITELIST

y-yagi:
### Summary

This is a bug fix. `linked_dirs` and `linked_files` are valid settings, but `doctor` task says that they are "not a recognized Capistrano setting".

```
    :linked_dirs is not a recognized Capistrano setting (/home/y-yagi/.rbenv/versions/2.7.2/lib/ruby/2.7.0/delegate.rb:83)
    :linked_files is not a recognized Capistrano setting (/home/y-yagi/.rbenv/versions/2.7.2/lib/ruby/2.7.0/delegate.rb:83
```

This fixed that issue.

### Short checklist

### Other Information

None

### DIFF
--- a/lib/capistrano/doctor/variables_doctor.rb
+++ b/lib/capistrano/doctor/variables_doctor.rb
@@ -10,6 +10,8 @@ module Capistrano
       WHITELIST = %i(
         application
         current_directory
+        linked_dirs
+        linked_files
         releases_directory
         repo_url
         repo_tree


y-yagi:
`whitelist` is tested by https://github.com/capistrano/capistrano/blob/5fe49bd6ea388b4dba19c658e8ba19d05802945f/spec/lib/capistrano/doctor/variables_doctor_spec.rb#L61-L65. Should I add tests of `whitelist` all values?

mattbrictson (approved):
Good catch! Thanks for the fix.

The reason we haven't seen this problem is that Capistrano recommends using `append` with these settings. E.g.

```ruby
append :linked_files, "config/database.yml"
append :linked_dirs, "log", "tmp/pids", "tmp/cache", "tmp/sockets", "public/system"
```

If you use `append` per the recommendation, you won't see the warnings from `doctor`.

However, I think `set` should still be allowed without warnings. So this PR is a good addition.

Also, I think this is good as-is without any additional tests.

Issue #1822, opened 2016-12-20
run_locally ignores --dry-run


BenoitHiller:
run_locally ignores the --dry-run flag and runs anyway. This seems to be because in https://github.com/capistrano/capistrano/blob/master/lib/capistrano/dsl.rb#L68 the current sshkit backend is ignored and the local one is used directly.

### Steps to reproduce
So with the following contrived test case:
```
namespace :test_case do
  desc 'A test for run_locally dry-run issue'
  task :testing do
    on roles(:app) do |host|
      run_locally do
        execute :echo, "fail"
      end
    end
  end
end
```
Then running it with `cap --dry-run stage test_case:testing`
### Expected behavior
It should output:
```
00:00 test_case:testing
      01 echo fail
```
Like it does if you remove the run_locally block.

### Actual behavior
It outputs:
```
00:00 test_case:testing
      01 echo fail
      01 fail
    ✔ 01 notroot@localhost 0.002s
```
Indicating it actually ran the program.

### System configuration
doctor output: https://gist.github.com/BenoitHiller/fc0fd11b95bf4faa5f7cd34da34d0711

leehambley:
`run_locally` is almost certainly going to go away. To help us weigh the ideal replacement for it, what exactly is your use-case for it?

We know that many people use it for tarring up files locally (a Rake `file` or `dir` task is better suited), or for building assets/etc using some build pipeline tool (for which Rake+Ruby's `sh` is better suited). When those replacements are used, the need to `run_locally` is avoided, and they all respect [the dry run flag from Rake][dry-run]. 

For some context, Capistrano's entire DSL is a subclassed `Rake::Application`, which allows us to stand on the shoulders of a giant, and have a consistent feeling, well-built robust API.

[dry-run]: http://rake.rubyforge.org/doc/command_line_usage_rdoc.html

BenoitHiller:
I'm honestly just using it to implement something similar to https://github.com/yannlugrin/capistrano-git-push but as a SCM plugin(see [this line](https://github.com/yannlugrin/capistrano-git-push/blob/master/lib/capistrano/git/push_strategy.rb#L29) for the basic idea). So probably not the kind of usage you are planning to support/condone.

However it looks like due to what is discussed in #1566 it isn't actually the case that capistrano does a Rake dry-run anymore, so `sh` isn't really a great solution.

[...]

PR #2133, opened 2023-05-31
Make Capistrano respect dry_run in run_locally. Fixes [capistrano#1822]

edspc:
### Summary

As there is no solution for issue #1822, I suggest updating `run_locally` to respect dry_run and adding `run_locally!` to ignore it and keep original behaviour.

Based on @tycooon [answer](https://github.com/capistrano/capistrano/issues/1822#issuecomment-663598976).

### Short checklist

### Other Information

I need this to use a custom SCM plugin that manages the git repository locally and creates a tar archive also locally. So during dry_run, I don't need to create an archive, but some git commands still should be executed.


mattbrictson:
@edspc thanks for the PR!

I don't see any unit tests so I assume you have been testing this manually. Can you describe your approach to testing this code, and some samples of the dry-run output you get? I'd like to work with you to get some automated test coverage in place for this behavior.

Also, we should update the documentation so people know about this behavior. Is there a place in the README or at <https://capistranorb.com> where it makes sense to mention it?

edspc:
Hi @mattbrictson!

I've added some tests and the notice to the docs.
Please feel free to update them if necessary)

mattbrictson:
@edspc Thanks! The structure of the tests looks good. I may make some tweaks before merging, but this is very helpful.

mattbrictson:
@edspc I redid the specs to remove the mocks while hopefully still testing what's important. What do you think?

PR #2135, opened 2023-06-13
Fix capistrano 2.x on Ruby 3 due to removed File.exists? method

mackuba:
I will not be surprised if you just close this 😅 but I saw that you've recently made some compatibility fixes and a bugfix release, so I thought it was worth a try.

Yesterday I tried to run `capify` from Capistrano 2 in a project on Ruby 3.2 and I got an error because of the `File.exists?` method, which was apparently long deprecated and was removed in 3.0:

https://github.com/ruby/ruby/commit/bf97415c02b11a8949f715431aca9eeb6311add2

This PR just replaces those few places in `capify` and elsewhere (+ tests) where that removed version of the method was used.

### DIFF
--- a/bin/capify
+++ b/bin/capify
@@ -22,7 +22,7 @@ end
 
 if ARGV.empty?
   abort "Please specify the directory to capify, e.g. `#{File.basename($0)} .'"
-elsif !File.exists?(ARGV.first)
+elsif !File.exist?(ARGV.first)
   abort "`#{ARGV.first}' does not exist."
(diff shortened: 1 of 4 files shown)


mattbrictson:
Looks good! Thanks for the fix 👍

mattbrictson:
I'll push a release for this in the next few days

mattbrictson:
Published to rubygems in [capistrano 2.15.11](https://rubygems.org/gems/capistrano/versions/2.15.11).

PR #2139, opened 2023-09-11
Remove default webpacker path to linked_dirs

taketo1113:
### Summary

Remove default webpacker path to linked_dirs.
Webpacker deprecated from rails v7.0.

https://github.com/rails/webpacker

### Short checklist

### DIFF
--- a/lib/capistrano/templates/deploy.rb.erb
+++ b/lib/capistrano/templates/deploy.rb.erb
@@ -24,7 +24,7 @@ set :repo_url, "git@example.com:me/my_repo.git"
 # append :linked_files, "config/database.yml", 'config/master.key'
 
 # Default value for linked_dirs is []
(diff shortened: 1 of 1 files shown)


mattbrictson (approved):
Thanks for the contribution! Some thoughts:

- Webpacker has been deprecated, but it is still widely used by existing Rails apps. However, this change only affects new apps (i.e. when `cap install` is run), so the potential negative impact to existing Capistrano-enabled Rails projects is minimal.
- Webpack support for Rails lives on in a gem called [shakapacker](https://github.com/shakacode/shakapacker). However, this gem uses a different cache location (`tmp/shakapacker`[^1]), so it will not be affected by this change.

Overall I think this is a good removal of cruft and should be safe. @leehambley @will-in-wi any objections?

[^1]: https://github.com/shakacode/shakapacker/blob/732ba010e6b54697875a19ad01e4975941a02c3f/lib/install/config/shakapacker.yml#L24

will-in-wi (approved):

leehambley:
Looks great, thanks for taking the initiative everyone 🙌

PR #2148, opened 2024-01-15
Fix RuboCop in CI

mattbrictson:
RuboCop in CI is now running on Ruby 3.3.0. This version no longer includes `racc` as a default gem, so the old version of RuboCop that we are using fails with this error:

> LoadError: cannot load such file -- racc/parser (LoadError)

Fix by explicitly adding `racc` to our Gemfile.

I also added the `base64` gem to address a deprecation warning.


mattbrictson:
Unrelated to this PR, the feature specs are failing due to problems with VirtualBox being unreliable on the latest GitHub Actions runners. I don't think it should block merging this PR.

PR #2155, opened 2024-05-19
Create `REVISION_TIME` with git commit timestamp upon deploy

### LINKED ISSUE #2153: Support exposing revision time alongside revision sha (when using git)
**Important:** GitHub issues are for feature requests or bug reports. The Capistrano team recommends you use [Stack Overflow](http://stackoverflow.com/questions/tagged/capistrano) for general questions. For more details, please see our [contribution policy](https://github.com/capistrano/capistrano/blob/master/CONTRIBUTING.md).

---

A while ago we had a request from a client that we have a light-weight "versioning" field in the admin panel of their app so they could check if there had been a new deployment - since we use continuous integration and deploy to autoblanced servers we couldn't use the release time as that is effectively arbitrary and not consistent (e.g. if we don't deploy for two weeks and a server is reprovisioned, it'll have a very different release time to the other servers), so to facilitate this we implemented a `REVISION_TIME` file into our Capistrano deployment which works pretty much exactly how the existing `REVISION` file does, just with Time™.

We saw this as useful to have in general so landed it in our `rails-template` (https://github.com/ackama/rails-template/pull/461) and have ported it to a number of our apps without issue.

This is something that feels like a useful and cheap win to be worth having upstreamed into Capistrano itself, especially since most of the code in that PR is just copied from what already exists in Capistrano since (afaik) there's no way to hook in any "closer".

G-Rath:
### Summary

The general story behind this is outlined in #2153 - ultimately, this has Capistrano start to manage a `REVISION_TIME` file which holds a unix timestamp of when the release was committed, in the same way that it stores that releases SHA in `REVISION`.

Resolves #2153

### Short checklist

### Other Information

I think this is actually pretty close to being landable since I figured out the general "how" a while ago in another project, but I expect some tweaking will probably be required and it would be nice to support this for svn and hg if possible but I don't use those tools so will either need someone who does to provide insight or I'll see if I can google/chatgpt a solution.

Some general questions I have are:
  - do we want to have a dedicated set of tasks, or fold this into the current "revision" based ones?
  - in the context of custom SCMs, do we want to make this an optional extra or required?
  - if optional, do we need to have any custom/extra checks? (I noticed that there are some `NotImplementedError` methods but it didn't seem like they were actually being called by Capistrano itself)
  - if the revision time isn't available, do we want to create an `REVISION_TIME` file at all? (it'd be simpler, but an empty file)

### DIFF
--- a/features/deploy.feature
+++ b/features/deploy.feature
@@ -40,6 +40,11 @@ Feature: Deploy
     Then the repo is cloned
     And the release is created
 
+  Scenario: REVISION and REVISION_TIME files are present
+    Given I make 1 deployment
+    And the REVISION file is created in the release
+    And the REVISION_TIME file is created in the release
+
   Scenario: Symlink linked files
     When I run cap "deploy:symlink:linked_files deploy:symlink:release" as part of a release
     Then file symlinks are created in the new release
@@ -85,4 +90,3 @@ Feature: Deploy
     Given I make 3 deployments
     When I rollback to a specific release
     Then the current symlink points to that specific release
-
--- a/features/deploy.feature
+++ b/features/deploy.feature
@@ -41,9 +41,9 @@ Feature: Deploy
     And the release is created
 
   Scenario: REVISION and REVISION_TIME files are present
-    Given I make 1 deployment
-    And the REVISION file is created in the release
-    And the REVISION_TIME file is created in the release
+    When I make 1 deployment
+    Then the REVISION file is created in the release
+    Then the REVISION_TIME file is created in the release
 
   Scenario: Symlink linked files
     When I run cap "deploy:symlink:linked_files deploy:symlink:release" as part of a release
--- a/features/step_definitions/assertions.rb
+++ b/features/step_definitions/assertions.rb
@@ -51,15 +51,13 @@ end
 Then(/^the REVISION file is created in the release$/) do
   stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION")
 
-  # TODO: confirm that this value is actually constant for everyone
-  expect(stdout.strip).to eq "3f41365139a0b3a408e96f13c781d9cedba8b4b8"
+  expect(stdout.strip).to match(/\h{40}/)
 end
 
 Then(/^the REVISION_TIME file is created in the release$/) do
   stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION_TIME")
 
-  # TODO: confirm that this value is actually constant for everyone
-  expect(stdout.strip).to eq "1718251930"
+  expect(stdout.strip).to match(/\d{10}/)
 end
 
 Then(/^file symlinks are created in the new release$/) do
--- a/features/step_definitions/assertions.rb
+++ b/features/step_definitions/assertions.rb
@@ -48,6 +48,20 @@ Then(/^the release is created$/) do
   run_remote_ssh_command("ls -g #{TestApp.releases_path}")
 end
 
+Then(/^the REVISION file is created in the release$/) do
+  stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION")
+
+  # TODO: confirm that this value is actually constant for everyone
+  expect(stdout.strip).to eq "3f41365139a0b3a408e96f13c781d9cedba8b4b8"
+end
+
+Then(/^the REVISION_TIME file is created in the release$/) do
+  stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION_TIME")
+
+  # TODO: confirm that this value is actually constant for everyone
+  expect(stdout.strip).to eq "1718251930"
+end
+
 Then(/^file symlinks are created in the new release$/) do
   TestApp.linked_files.each do |file|
     run_remote_ssh_command(test_symlink_exists(TestApp.current_path.join(file)))
--- a/lib/capistrano/scm/tasks/git.rake
+++ b/lib/capistrano/scm/tasks/git.rake
@@ -70,4 +70,15 @@ namespace :git do
       end
     end
   end
+
+  desc "Determine the unix timestamp that the revision that will be deployed was created"
+  task :set_current_revision_time do
+    on release_roles(:all), in: :groups, limit: fetch(:git_max_concurrent_connections), wait: fetch(:git_wait_interval) do
+      within repo_path do
+        with fetch(:git_environmental_variables) do
+          set :current_revision, git_plugin.fetch_revision
+        end
+      end
+    end
+  end
 end
--- a/lib/capistrano/scm/tasks/git.rake
+++ b/lib/capistrano/scm/tasks/git.rake
@@ -76,7 +76,7 @@ namespace :git do
     on release_roles(:all), in: :groups, limit: fetch(:git_max_concurrent_connections), wait: fetch(:git_wait_interval) do
       within repo_path do
         with fetch(:git_environmental_variables) do
-          set :current_revision, git_plugin.fetch_revision
+          set :current_revision_time, git_plugin.fetch_revision_time
         end
       end
     end
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -259,7 +259,9 @@ namespace :deploy do
   task :set_current_revision_time do
     on release_roles(:all) do
       within release_path do
-        execute :echo, "\"#{fetch(:current_revision_time)}\" > REVISION_TIME"
+        if fetch(:current_revision_time)
+          execute :echo, "\"#{fetch(:current_revision_time)}\" > REVISION_TIME"
+        end
       end
     end
   end
--- a/lib/capistrano/tasks/deploy.rake
+++ b/lib/capistrano/tasks/deploy.rake
@@ -3,6 +3,7 @@ namespace :deploy do
     invoke "deploy:print_config_variables" if fetch(:print_config_variables, false)
     invoke "deploy:check"
     invoke "deploy:set_previous_revision"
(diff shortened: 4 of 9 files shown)


will-in-wi (approved):

lib/capistrano/scm/tasks/git.rake
```diff
@@ -70,4 +70,15 @@ namespace :git do
       end
     end
   end
+
+  desc "Determine the unix timestamp that the revision that will be deployed was created"
+  task :set_current_revision_time do
+    on release_roles(:all), in: :groups, limit: fetch(:git_max_concurrent_connections), wait: fetch(:git_wait_interval) do
+      within repo_path do
+        with fetch(:git_environmental_variables) do
+          set :current_revision, git_plugin.fetch_revision
```
mattbrictson, line 79: ❓ I think there might be a typo here. Shouldn't it be:

```suggestion
          set :current_revision, git_plugin.fetch_revision_time
```

lib/capistrano/tasks/deploy.rake
```diff
@@ -253,6 +255,24 @@ namespace :deploy do
     end
   end
 
+  desc "Place a REVISION_TIME file with the current revision commit time in the current release path"
+  task :set_current_revision_time do
+    on release_roles(:all) do
+      within release_path do
+        execute :echo, "\"#{fetch(:current_revision_time)}\" > REVISION_TIME"
```
mattbrictson, line 262: ❓ Should this be skipped if `fetch(:current_revision_time)` is nil?

spec/lib/capistrano/scm/git_spec.rb
```diff
@@ -170,6 +171,15 @@ module Capistrano
       end
     end
 
+    describe "#fetch_revision_time" do
```
mattbrictson, line 174: ✨ Thanks for the spec!

mattbrictson (changes requested):
Thanks for the PR! To answer your questions:

> do we want to have a dedicated set of tasks, or fold this into the current "revision" based ones?

I slightly prefer the dedicated tasks.

> in the context of custom SCMs, do we want to make this an optional extra or required?

For now, I think it can be optional. I personally don't have much recent hg or svn experience, so I hesitate on making it required for those other SCMs. It looks like the only time the `fetch_revision_time` method is called is by the `deploy:set_current_revision_time` before-hook that the git plugin registers, so if that method isn't implemented by the other plugins, it shouldn't be a problem.

> if the revision time isn't available, do we want to create an REVISION_TIME file at all? (it'd be simpler, but an empty file)

I think we should skip creating the REVISION_TIME file in this case (see inline comment below).

Finally, I think it would be nice to add a feature (cucumber) test for this. Maybe just a simple assertion that the REVISION_TIME file is created upon successful deploy. I'll be merging #2159 soon, which will make it easier to run the feature tests.

lib/capistrano/tasks/deploy.rake
G-Rath: I don't mind making this conditional, but just skipping entirely would mean that if the revision time couldn't be gotten for some reason in this deployment, the previous revision time would be retained whereas piping an empty string to the file will clear it:

```
~ via 🐍 v3.10.12 via 💎
❯ echo 'hello world' > file

~ via 🐍 v3.10.12 via 💎
❯ cat file
hello world

~ via 🐍 v3.10.12 via 💎
❯ echo '' > file

~ via 🐍 v3.10.12 via 💎
❯ cat file

```

While in theory it should always work or not work, I think it's nice if we have some just-in-case logic for this given it should be cheap - with that in mind do you have a preference for:
  - always outputting to the `REVISION_TIME` file, or
  - always remove the `REVISION_TIME` file first then echo conditionally, or
  - remove the `REVISION_TIME` file if it exists then do the echo conditionally

lib/capistrano/scm/tasks/git.rake
G-Rath: whoops yeah good catch

G-Rath:
@mattbrictson thanks for the review! I'm assuming based on the comments (or lack thereof) on the PR that my changes already match the opinions you've expressed in reply to my questions, but let me know if that's not the case!

In the meantime I'll checkout #2159 and see if I can get a feature test crafted

lib/capistrano/tasks/deploy.rake
mattbrictson: I'm not sure I understand this part:

> if the revision time couldn't be gotten for some reason in this deployment, the previous revision time would be retained

When `cap deploy` runs, it creates a new release directory, so if nothing else happens, then there will be no `REVISION_TIME` present. In what circumstances would the previous release's `REVISION_TIME` file end up in the newly created release?

Regardless, I think what I'm looking for is for the `REVISION_TIME` file to be absent if something goes wrong, or if a non-supported SCM is being used. I'm worried about Capistrano adding empty files into users' releases that serve no apparent purpose.

G-Rath: > When cap deploy runs, it creates a new release directory

Yeah sorry I swear it's being a day and a half today... I was thinking of the `revisions.log` which is kept in the root directory whereas `REVISION` (and now `REVISION_TIME`) are kept in the directory for each release, which makes tons of sense because of this very situation

G-Rath:
@mattbrictson I'm about to knock off but hopefully will pick this up tomorrow - in the meantime if you're around, I wouldn't mind hearing what you think about the "then the release path is created" assertion:

https://github.com/capistrano/capistrano/blob/3f41365139a0b3a408e96f13c781d9cedba8b4b8/features/step_definitions/assertions.rb#L18-L20

Which is used in

https://github.com/capistrano/capistrano/blob/3f41365139a0b3a408e96f13c781d9cedba8b4b8/features/deploy.feature#L37-L41

I was thinking of adding to that spec or using it to create a new one, but while looking at it it doesn't seem like it actually asserts anything? it's doing `ls -g` of the `releases` path rather than an actual release, so if I understand correctly it will only error if the `releases` directory doesn't exist - is that intentional?

features/step_definitions/assertions.rb
```diff
@@ -6,67 +6,77 @@
...
 
 Then(/^the release is created$/) do
-  run_vagrant_command("ls -g #{TestApp.releases_path}")
+  run_remote_ssh_command("ls -g #{TestApp.releases_path}")
+end
+
+Then(/^the REVISION file is created in the release$/) do
+  stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION")
+
+  # TODO: confirm that this value is actually constant for everyone
```
G-Rath, line 54: give the specs are passing, it seems like these are consistent though I'm still not sure where they're coming from exactly - either way, I'll leave these TODOs in at least until #2159 is landed

features/deploy.feature
```diff
@@ -40,6 +40,11 @@ Feature: Deploy
     Then the repo is cloned
     And the release is created
 
+  Scenario: REVISION and REVISION_TIME files are present
+    Given I make 1 deployment
+    And the REVISION file is created in the release
+    And the REVISION_TIME file is created in the release
```
mattbrictson, line 46: I think this would read more naturally:

```suggestion
    When I make 1 deployment
    Then the REVISION file is created in the release
    Then the REVISION_TIME file is created in the release
```

features/step_definitions/assertions.rb
```diff
@@ -6,67 +6,77 @@
...
+  expect(stdout.strip).to eq "3f41365139a0b3a408e96f13c781d9cedba8b4b8"
```
mattbrictson, line 55: I think it would be better to match this with a regular expression, because I doubt this value is going to be stable. How about:

```suggestion
  expect(stdout.strip).to match(/\h{40}/)
```

features/step_definitions/assertions.rb
```diff
@@ -6,67 +6,77 @@
...
+end
+
+Then(/^the REVISION_TIME file is created in the release$/) do
+  stdout, _stderr = run_remote_ssh_command("cat #{@release_paths[0]}/REVISION_TIME")
+
+  # TODO: confirm that this value is actually constant for everyone
+  expect(stdout.strip).to eq "1718251930"
```
mattbrictson, line 62: Likewise,

```suggestion
  expect(stdout.strip).to match(/\d{10}/)
```

mattbrictson (approved):

mattbrictson:
> it's doing `ls -g` of the releases path rather than an actual release, so if I understand correctly it will only error if the releases directory doesn't exist - is that intentional?

@G-Rath I think you're right. Do you mind opening a GitHub issue for this so we can track it as a bug?

G-Rath:
@mattbrictson rather than an issue, I've opened a pull request: https://github.com/capistrano/capistrano/pull/2160

mattbrictson:
@G-Rath if for whatever reason, someone wanted to disable this new `REVISION_TIME` file from being created, how would they do that? I want to document it in the release notes, just to be safe.

G-Rath:
@mattbrictson the sure fire way would be to ensure that `fetch(:current_revision_time)` returns a false-y value, which could be done by registering a task straight after the `deploy:set_current_revision_time` task that explicitly sets it to `nil`.

I've not done it myself but I assume they could also remove or override one of the tasks in the chain, such as `deploy:set_previous_revision_time`

G-Rath:
Again, not done it myself but [this is what I was thinking of](https://capistranorb.com/documentation/advanced-features/overriding-capistrano-tasks/) - that should be enough right? combined with the task name of `deploy:set_current_revision_time`

mattbrictson:
@G-Rath yes, that makes sense. I'll test that out.

mattbrictson:
This seems to do the trick:

```ruby
Rake::Task["git:set_current_revision_time"].clear_actions
```

mattbrictson:
This feature has been released in Capistrano 3.19.0: https://github.com/capistrano/capistrano/releases/tag/v3.19.0

PR #2159, opened 2024-06-11
Use Docker container as remote SSH target for test suite

eric-eye:
### Summary

Capistrano's Cucumber suite is expected to run against Vagrant, which is not fully supported on Silicon Macs and produces friction in CI.

This PR swaps the remote SSH target to use a Docker container, which allows for the Cucumber suite to be runnable on Silicon Macs and in CI. To run the test suite, the Docker daemon must be running first.

### Short checklist

- [x] ~If relevant, did you create a test?~ (none added, but some are modified)

### Other Information

Developed with @mattbrictson and @bunnymatic


bunnymatic:
FWIW - I'm running `colima`

Ran 
```
colima start
bundle exec cucumber
```
and saw
```
28 scenarios (1 pending, 27 passed)
152 steps (1 pending, 151 passed)
2m13.597s
```

bunnymatic (approved):

.docker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:22.04
```
mattbrictson, line 1: 🗒️ This is the Dockerfile that defines a container running an sshd server. We use it as the deployment target for the Cucumber tests.

.docker/ssh_key_rsa
```diff
@@ -0,0 +1,49 @@
+-----BEGIN OPENSSH PRIVATE KEY-----
```
mattbrictson, line 1: 🗒️ This is the key used to authenticate SSH connections to the ephemeral Docker container during Cucumber tests. This key is used for no other purpose, so there is no problem checking it into the Git repo.

.github/workflows/ci.yml
```diff
@@ -68,16 +68,9 @@ jobs:
         run: bundle exec rake rubocop
   features:
     needs: [spec, spec-legacy]
-    runs-on: macos-12
+    runs-on: ubuntu-latest
```
mattbrictson, line 71: 🗒️ Previously we had to run the Cucumber job on a macOS runner. This was a hack because macOS was the only runner able to run VirtualBox, and in the past year, this has become less and less reliable, to the point where VirtualBox would almost always hang or fail to start.

Now that we are using the Docker image to run the sshd container instead of VirtualBox, we can use a standard Ubuntu runner, which includes reliable Docker support.

spec/support/test_app.rb
```diff
@@ -96,13 +100,12 @@ def cap(task, subdirectory=nil)
   end
 
   def run(command, subdirectory=nil)
-    output = nil
     command = "bundle exec #{command}" unless command =~ /^bundle\b/
     dir = subdirectory ? test_app_path.join(subdirectory) : test_app_path
-    Dir.chdir(dir) do
-      output = with_clean_bundler_env { `#{command}` }
+    output, status = Dir.chdir(dir) do
+      with_clean_bundler_env { Open3.capture2e(command) }
```
mattbrictson, line 106: 🗒️ Changed from backticks to `capture2e` so that stderr from the `cap` command doesn't clutter the console when running the Cucumber tests.

G-Rath:
fwiw I'm on WSLv2 with Docker-for-Windows, and this Just Worked 🎉

It did seem like it was hung on the first spec, because it was building the docker container which took a minutes and there was no output until it was done - I think if it's very easy I'd try to do the build in like a global setup hook where you can say "building docker containers", but since pulling + building should be an infrequent action, I totally get not doing that if its more than 5 minutes work

(the full suite btw took 2 minutes for me)

mattbrictson:
@G-Rath thanks for trying this out!

I agree that the first-run experience is a bit confusing, given the delay with no immediate output. It's important to me to be accommodating to new contributors, so I think this is worth fixing.

features/support/docker_gateway.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def run_shell_command(command)
+    run_compose_command("exec ssh_server /bin/bash -c #{command.shellescape}")
+  end
+
+  private
+
+  def run_compose_command(command)
+    log "[docker compose] #{command}"
+    stdout, stderr, status = Open3.capture3("docker compose #{command}")
```
mattbrictson, line 28: ❓ Maybe we should stream the output while capturing it? `Open3.capture3` blocks until the process completes, so if the command is slow (e.g. downloading the docker base image), there won't be any output printed, and the test suite may appear "stuck" as a result.

Perhaps using a technique like this: https://mattbrictson.com/blog/run-shell-commands-in-ruby#streaming-output-while-capturing-it

eric-eye: Thanks for the feedback, @G-Rath!

@mattbrictson I implemented something similar to your blog post as a new commit, but it's not quite seamless with the Cucumber logger. Two thing I noticed:

- Cucumber logger does not seem to flush messages as soon as they are received, and I suspect this is because it uses an STDOUT i/o whose `#sync == false`
- For reasons I don't grok, messages for image building (which can be numerous) and routine messages from docker-compose seem to be emitted differently, so `popen3` seemingly does not stream the latter (so we also continue the "capture then print" approach as before)

So I think this solves the radio silence issue, but symmetry with our Cucumber logger is still missing.

mattbrictson: @eric-eye ah, I see. Thanks for looking into that. My interpretation is that Cucumber does not print the `log` output until the test step finishes, which prevents us from streaming the output through that mechanism.

https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/formatter/pretty.rb#L116

I think your commit is a good compromise for now to fix the "radio silence" issue. Thanks!

mattbrictson (approved):

PR #2160, opened 2024-06-18
Properly assert that release was created in feature test

G-Rath:
### Summary

Follow up to https://github.com/capistrano/capistrano/pull/2155#issuecomment-2172283634 and https://github.com/capistrano/capistrano/pull/2155#issuecomment-2174466232

I've gone with expecting the year, month, day as for that to fail there'd need to be a 24 hour delay between steps which'd be a whole problem of its own - it might be possible that if you run the suite at midnight in a particular timezone it could also fail, but if so that should only be a ~30 second window of possibility...

### Short checklist

### Other Information

This is branched off #2159

### DIFF
--- a/features/step_definitions/assertions.rb
+++ b/features/step_definitions/assertions.rb
@@ -45,7 +45,9 @@ Then(/^the repo is cloned$/) do
 end
 
 Then(/^the release is created$/) do
-  run_remote_ssh_command("ls -g #{TestApp.releases_path}")
+  stdout, _stderr = run_remote_ssh_command("ls #{TestApp.releases_path}")
+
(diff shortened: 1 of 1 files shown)


mattbrictson:
@G-Rath looks good, thanks! 🙇  It would be great to avoid the flakiness around midnight, but I can't think of a way to do that without adding a significant amount of complexity. As it stands, the flake scenario should very rarely happen.

mattbrictson:
@G-Rath wow, you're quick! Thanks for the rebase

mattbrictson (approved):
Nice! ✨

PR #2167, opened 2025-05-08
fix: update RuboCop config

paulschreiber:
### Summary
The RuboCop config is using outdated cop names. Running it with 3.19.2 produces the following warnings:
```
vendor/bundle/ruby/3.3.0/gems/capistrano-3.19.2/.rubocop.yml: Style/SpaceAroundEqualsInParameterDefault has the wrong namespace - replace it with Layout/SpaceAroundEqualsInParameterDefault
Error: The `Metrics/LineLength` cop has been moved to `Layout/LineLength`.
(obsolete configuration found in vendor/bundle/ruby/3.3.0/gems/capistrano-3.19.2/.rubocop.yml, please update it)
The `Style/AccessorMethodName` cop has been moved to `Naming/AccessorMethodName`.
(obsolete configuration found in vendor/bundle/ruby/3.3.0/gems/capistrano-3.19.2/.rubocop.yml, please update it)
The `Style/BracesAroundHashParameters` cop has been removed. Please use  and/or  instead.
(obsolete configuration found in vendor/bundle/ruby/3.3.0/gems/capistrano-3.19.2/.rubocop.yml, please update it)
```

This updates the cop names and eliminates the warnings.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,16 +13,13 @@ Metrics/BlockLength:
     - "lib/**/*.rake"
 Style/BarePercentLiterals:
   EnforcedStyle: percent_q
-Style/BracesAroundHashParameters:
-  Exclude:
-    - spec/integration/dsl_spec.rb
 Style/ClassAndModuleChildren:
   Enabled: false
 Style/DoubleNegation:
   Enabled: false
 Style/IndentHeredoc:
   Enabled: false
-Style/SpaceAroundEqualsInParameterDefault:
+Layout/SpaceAroundEqualsInParameterDefault:
   EnforcedStyle: no_space
 Style/StringLiterals:
   EnforcedStyle: double_quotes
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks! Unfortunately I can't accept this change at this time, because we are stuck running an old version of RuboCop.

Capistrano supports very old versions of Ruby, all the way back to Ruby 2.0. That means that we have to run a very old version of RuboCop (0.48.1); newer versions of RuboCop do not support Ruby 2.0. Which is why we are using old cop names that trigger warnings on newer versions.

It would be great if we could support the latest version of RuboCop and also guarantee that our code is Ruby 2.0 compatible, but I don't know if that's possible. For now we're sticking with the old version of RuboCop.

If you have any suggests on how we might improve our RuboCop setup, let me know!

